PR #35, opened 2020-03-23
[helm] Add support for using NATS Account Server as the resolver

wallyqs:
This adds a Helm chart for the account server and support for the NATS Helm chart to use the same secrets.  Flow to use both charts would like this:

- First, create an NSC setup locally 

```sh
export NKEYS_PATH=$(pwd)/nsc/nkeys
export NSC_HOME=$(pwd)/nsc/accounts
curl -sSL https://nats-io.github.io/k8s/setup/nsc-setup.sh | sh
nsc edit operator --account-jwt-server-url http://nats-account-server:9090/jwt/v1/ --service-url nats://nats:4222
```

- Upload the shared secrets and configmaps

```sh
kubectl create configmap operator-jwt --from-file ./nsc/accounts/nats/KO/KO.jwt
kubectl create configmap nats-sys-jwt --from-file ./nsc/accounts/nats/KO/accounts/SYS/SYS.jwt
kubectl create secret generic nats-sys-creds --from-file ./nsc/nkeys/creds/KO/SYS/sys.creds
```

- Create the Helm deploy yaml for the NATS Server
```sh
echo '
# NATS Server connection settings.
nats:
  # NATS Service to which we can connect.
  url: "nats://nats:4222"

  # Credentials to connect to the NATS Server.
  credentials:
    secret:
      name: nats-sys-creds
      key: sys.creds

# Trusted Operator mode settings.
operator:
  # Reference to the system account jwt.
  systemaccountjwt:
    configMap:
      name: nats-sys-jwt
      key: SYS.jwt

  # Reference to the Operator JWT.
  operatorjwt:
    configMap:
      name: operator-jwt
      key: KO.jwt
' > deploy-account-server.yaml
```

- Create the NATS Account Server with file storage and a persistent volume

```sh
helm install nats-account-server -f deploy-account-server.yaml ./helm/charts/nats-account-server/
```
- Now need to first feed into the NATS Account Server the current set of accounts
  NOTE: This requires to `sudo vi /etc/hosts` to add localhost temporarily so that it resolves 127.0.0.1 as the K8S service domain.

```
# 
# 127.0.0.1 nats-account-server
# 
```

- Open the port for the nats-account-server so that can feed the accounts locally

```sh
kubectl port-forward nats-account-server-0 9090:9090 &
nsc push -A
```

- Get the public key of system account from the NATS server

```sh
nsc list accounts
╭─────────────────────────────────────────────────────────────────╮
│                            Accounts                             │
├──────┬──────────────────────────────────────────────────────────┤
│ Name │ Public Key                                               │
├──────┼──────────────────────────────────────────────────────────┤
│ A    │ ADR5DUK5TCFGAUHQ76QP4V7GXIEBYTQ26STXCRTP4X2BJDMUMSBXWPWS │
│ B    │ ABIDLGSR6YZBDGH3C4FK6WI6MMVR7U7M6MR3RI3GKEUCJD27UFHMOKY5 │
│ STAN │ AAF5E33YN7DIKFQVR3R52XOM5GOM7SIILPITZXW3DFVU743QICIN3Z3J │
│ SYS  │ ADHUAQZ7UEXIVAXNI6B4VUU4IBV2ZUXMQMSIYDFQQ5BFTSMWTKK2NWJN │ < this one
╰──────┴──────────────────────────────────────────────────────────╯
```

```sh
# Helm config for NATS
echo '
# Authentication setup
auth:
  enabled: true

  # Reference to the Operator JWT which will be mounted as a volume,
  # shared with the account server in this case.
  operatorjwt:
    configMap:
      name: operator-jwt
      key: KO.jwt

  # Public key of the System Account
  systemAccount: ADHUAQZ7UEXIVAXNI6B4VUU4IBV2ZUXMQMSIYDFQQ5BFTSMWTKK2NWJN

  resolver:
    type: URL
    url: "http://nats-account-server:9090/jwt/v1/accounts/"

# Add system credentials to the nats-box instance for example
natsbox:
  enabled: true

  credentials:
    secret:
      name: nats-sys-creds
      key: sys.creds
' > deploy-nats.yaml
```

- Now deploy the NATS Server with the URL resolver configuration against the NATS Account Server running in the same cluster

```sh
helm install nats -f deploy-nats.yaml ./helm/charts/nats/
```

- Can use nats-box with the injected system credentials accounts now as well:

```sh
kubectl exec -n default -it nats-box -- /bin/sh -l
nats-sub -creds /etc/nats-config/syscreds/sys.creds '>'
nats-sub '>' # also works since $USER_CREDS env var is set automatically
```

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -76,6 +76,11 @@ natsbox:
   image: synadia/nats-box:0.3.0
   pullPolicy: IfNotPresent
 
+  # credentials:
+  #   secret:
+  #     name: nats-sys-creds
+  #     key: sys.creds
+
 # The NATS config reloader image to use.
 reloader:
   enabled: true
@@ -92,22 +97,34 @@ exporter:
 auth:
   enabled: false
 
-  # Operator JWT
-  operator: 
+  # Reference to the Operator JWT.
+  # operatorjwt:
+  #   configMap:
+  #     name: operator-jwt
+  #     key: KO.jwt
 
   # Public key of the System Account
-  systemAccount: 
+  # systemAccount:
 
-  resolver:
-    type: URL
-
-    # URL resolver settings
-    url: "http://nats-account-server:9090/jwt/v1/accounts/"
-
-    # 
-    # Use a configmap reference which will be mounted
-    # into the container.
-    # 
-    configMap:
-      name: nats-accounts
-      key: resolver.conf
+  # resolver:
+  #   ############################
+  #   #                          #
+  #   # Memory resolver settings #
+  #   #                          #
+  #   ##############################
+  #   # type: memory
+  #   # 
+  #   # Use a configmap reference which will be mounted
+  #   # into the container.
+  #   # 
+  #   # configMap:
+  #   #   name: nats-accounts
+  #   #   key: resolver.conf
+  #  
+  #   ##########################
+  #   #                        #
+  #   #  URL resolver settings #
+  #   #                        #
+  #   ##########################
+  #   # type: URL
+  #   # url: "http://nats-account-server:9090/jwt/v1/accounts/"
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -72,7 +72,7 @@ gateway:
   # List of remote gateways
   # gateways:
   #   - name: other
-  #     url: tls://my-gateway-url:7522
+  #     url: nats://my-gateway-url:7522
   
 # In case of both external access and advertisements being
 # enabled, an initializer container will be used to gather
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -91,8 +91,19 @@ exporter:
 # Authentication setup
 auth:
   enabled: false
+
+  # Operator JWT
+  operator: 
+
+  # Public key of the System Account
+  systemAccount: 
+
   resolver:
-    type: memory
+    type: URL
+
+    # URL resolver settings
+    url: "http://nats-account-server:9090/jwt/v1/accounts/"
+
     # 
     # Use a configmap reference which will be mounted
     # into the container.
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -1,6 +1,10 @@
-# NATS Server Configuration
+###############################
+#                             #
+#  NATS Server Configuration  #
+#                             #
+###############################
 nats:
-  image: nats:2.1.4-alpine3.11
+  image: nats:2.1.6-alpine3.11
   pullPolicy: IfNotPresent
 
   # Toggle whether to enable external access.
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -59,7 +59,21 @@ cluster:
 #
 leafnodes:
   enabled: false
+  # remotes:
+  #   - url: "tls://connect.ngs.global:7422"
 
+# Gateway connections to create a super cluster
+#
+# https://docs.nats.io/nats-server/configuration/gateways
+#
+gateway:
+  enabled: false
+  name: 'default'
+  # List of remote gateways
+  # gateways:
+  #   - name: other
+  #     url: tls://my-gateway-url:7522
+  
 # In case of both external access and advertisements being
 # enabled, an initializer container will be used to gather
 # the public ips.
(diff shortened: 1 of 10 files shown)


ColinSullivan1:
For the sub in nats-box, I had to use:  `nats-sub -s nats -creds /etc/nats-config/creds/sys.creds '>'`

ColinSullivan1:
If the issue below gets implemented then this could be much cleaner.

https://github.com/nats-io/nsc/issues/264

wallyqs:
agree, I think having that in `nsc` would help the guide for setting this up, although still have to manually create the config yaml with the system account since that is not controlled by helm.

ColinSullivan1:
Continuing to think of simplification...  What if the server config could reference a path to the system account jwt?

matthiashanel:
Use an environment variable to look up the account?

 acc=`nsc list accounts | grep "SYS[^A]*" | grep "A[A-Z0-9]*" -o`

matthiashanel:
where it possible to turn systemAccount: int a config map as well?
Then modifying the config wouldn't be necessary, which would simplify things. as well.
Same with http://nats-account-server:9090/jwt/v1/ I guess.

ColinSullivan1:
imo cleanest and easiest to use approach would be where the NATS server references the system jwt file from its config and looks up the public key.  Could be a new parameter, `systemAccountFile` (`system_file` in conf), which supports only the public JWT.  ...certainly open to other ideas.  @derekcollison wdyt?

ColinSullivan1:
Alternatively one could pass the key or a path to the system account JWT as a parameter in the NATS server.

ColinSullivan1:
For the config map solution do you mean to  put the system key in a config map then reference via env var?

https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-container-environment-variables-using-configmap-data

matthiashanel:
@ColinSullivan1 modifying the config seems like an error prone process. 
So why not have the configuration make more use of configMaps as the nats server does.
There every moving piece seems to be configurable externally.
In the account server config this is not the case.

I think it would be much cleaner to say this are the individual pieces necessary.
Together with this configuration (that you don't modify) you can start the service.

The comment about environment variable was more like store the account name in an environment variable then have echo using it.... but config map seems to be even better for this.

> Be aware, all of this is based on my current understanding of how helm charts work.

ColinSullivan1:
> @ColinSullivan1 modifying the config seems like an error prone process.
> So why not have the configuration make more use of configMaps as the nats server does.

I agree. The NATS server can reference an environment variable in it's config...   Then the deployment yaml can assign the configmap to that environment var.  But we still have the copy/paste step - just in a different spot.

wallyqs:
Thanks everyone for the ideas, there might be a way to avoid the copy/paste by using an extra configmap then have that be injected as an env variable.

```
kubectl create configmap literal-data --from-literal systemAccount="$(nsc list accounts | grep "SYS[^A]*" | grep "A[A-Z0-9]*" -o)" 
```

Though yes that would also be improved if `nsc` could make it easier to get this info.

matthiashanel:
@wallyqs , I have question regarding:

> nsc edit operator --account-jwt-server-url http://nats-account-server:9090/jwt/v1/ --service-url nats://nats:4222
Since you do this is it even still necessary to specify the following in # Helm config for NATS?:
> resolver:
>    type: URL
>    url: "http://nats-account-server:9090/jwt/v1/accounts/"

wallyqs:
@matthiashanel that is a good question, since the operator jwt already has the metadata related to the dns name of the location of the account server, maybe the nats server could use the url there to send the request, but it doesn't work like that right now... so yes would be needed.
I think the ideal scenario would be to be able to: 

- Use `nsc push --account-jwt-server-url $SOME_IP` to push the accounts to be able to push to localhost or any interface without re-modifying the operator jwt payload
- Embed the location of the account server in the operator jwt, then have the nats server use that without having declaring the `resolver: URL()`
- Have the system account setting in the nats server be able to use a file like the account server or make the nats account server also use a literal system account public nkey, so that this is consistent in both... then this could be shared as an extra config map

(One issue with too many configmaps and includes is that the overall config becomes a bit more complex as they end up volume declarations in the yaml which are more difficult to maintain...)

helm/charts/nats/values.yaml
```diff
@@ -91,12 +110,35 @@ exporter:
...
+  #   # Use a configmap reference which will be mounted
+  #   # into the container.
+  #   # 
+  #   # configMap:
+  #   #   name: nats-accounts
+  #   #   key: resolver.conf
+  #  
+  #   ##########################
+  #   #                        #
+  #   #  URL resolver settings #
```
matthiashanel, line 140: I would maybe mention that this is only needed if not part of the jwt anyway.

PR #36, opened 2020-03-24
[helm] Adding support for gateways and leafnodes

wallyqs:
Makes it possible to setup remote leafnodes and gateways as follows:

```yaml

# Authentication setup
auth:
  enabled: true

  # Reference to the Operator JWT which will be mounted as a volume,
  # shared with the account server in this case.
  operatorjwt:
    configMap:
      name: operator-jwt
      key: KO.jwt

  # Public key of the System Account
  systemAccount: AAITGVORQ4VHOQ32A7XMEKUIDMJ5GSAYSUGOV6GZAVSQWYFAL72DIXLC

  resolver:
    type: URL
    url: "http://nats-account-server:9090/jwt/v1/accounts/"

nats:
  externalAccess: true
  logging:
    debug: true
    trace: true

leafnodes:
  enabled: true
  remotes:
    - url: tls://connect.ngs.global:7422
      credentials:
        secret:
          name: ngs-creds
          key: NGS.creds

gateway:
  enabled: true
  name: aws-useast2
  gateways:
    - name: euwest1
      url: tls://euwest1.aws.ngs.global:7522

# Add system credentials to the nats-box instance for example
natsbox:
  enabled: true

  credentials:
    secret:
      name: nats-sys-creds
      key: sys.creds
```

```
kubectl create secret generic ngs-creds --from-file ./.nkeys/creds/synadia/NGS/NGS.creds
helm install nats -f deploy-nats-gateways.yaml ./helm/charts/nats/
```

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -53,7 +53,21 @@ cluster:
 #
 leafnodes:
   enabled: false
+  # remotes:
+  #   - url: "tls://connect.ngs.global:7422"
 
+# Gateway connections to create a super cluster
+#
+# https://docs.nats.io/nats-server/configuration/gateways
+#
+gateway:
+  enabled: false
+  name: 'default'
+  # List of remote gateways
+  # gateways:
+  #   - name: other
+  #     url: tls://my-gateway-url:7522
+  
 # In case of both external access and advertisements being
 # enabled, an initializer container will be used to gather
 # the public ips.
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -66,7 +66,7 @@ gateway:
   # List of remote gateways
   # gateways:
   #   - name: other
-  #     url: tls://my-gateway-url:7522
+  #     url: nats://my-gateway-url:7522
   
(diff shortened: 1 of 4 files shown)


helm/charts/nats/values.yaml
```diff
@@ -53,7 +53,21 @@ cluster:
...
+#
+# https://docs.nats.io/nats-server/configuration/gateways
+#
+gateway:
+  enabled: false
+  name: 'default'
+  # List of remote gateways
+  # gateways:
+  #   - name: other
+  #     url: tls://my-gateway-url:7522
```
matthiashanel: If I understand correctly what I found out yesterday, this should literally not work because the tls section is missing that contains the cert that is used. It fails because gateways will verify client connection and need a cert.

wallyqs: yes that's right, it is just an example. Have to do tls blocks on another pr but in current pr only plain text is supported

matthiashanel: For the time being, change tls//: to nats:// then?

wallyqs: yeah sounds good

PR #44, opened 2020-04-09
Addex max_channels config value

alexander-klang:


wallyqs:
Hi, thanks for the PR but you can already do this by using `limits`. 
https://github.com/nats-io/k8s/blob/master/helm/charts/stan/values.yaml#L52

```yaml
store:
  limits:
    max_channels: 1024
```

Sorry for the confusion.

alexander-klang:
That is great! Thank you for clarifying!

Issue #58, opened 2020-05-13
Helm: store.file.path does not seem to be working on NFS


samstride:
Hi,

I have attached an NFS to our k8s cluster at path `/var/mnt`

I have installed nats + stan using the latest helm charts (as of this writing) like this

```sh
helm install nats-server nats/nats --namespace=nats 

helm install stan-server nats/stan --namespace=nats \
  --set stan.nats.url=nats://nats-server:4222 \
  --set store.type=file \
  --set store.file.path=/var/mnt/nats \
  --set store.file.storageSize=1Gi
```

The startup log from stan shows the NFS path being set

```
[1] 2020/05/13 04:03:38.932852 [INF] STREAM: Starting nats-streaming-server[stan-server] version 0.17.0
[1] 2020/05/13 04:03:38.933013 [INF] STREAM: ServerID: xDec25eoUVPr0wxKKaj7xS
[1] 2020/05/13 04:03:38.933029 [INF] STREAM: Go version: go1.13.7
[1] 2020/05/13 04:03:38.933041 [INF] STREAM: Git commit: [f4b7190]
[1] 2020/05/13 04:03:38.943212 [INF] STREAM: Recovering the state...
[1] 2020/05/13 04:03:38.945227 [INF] STREAM: No recovered state
[1] 2020/05/13 04:03:39.196686 [INF] STREAM: Message store is FILE
[1] 2020/05/13 04:03:39.196697 [INF] STREAM: Store location: /var/mnt/nats
[1] 2020/05/13 04:03:39.196721 [INF] STREAM: ---------- Store Limits ----------
[1] 2020/05/13 04:03:39.196723 [INF] STREAM: Channels:                  100 *
[1] 2020/05/13 04:03:39.196725 [INF] STREAM: --------- Channels Limits --------
[1] 2020/05/13 04:03:39.196726 [INF] STREAM:   Subscriptions:          1000 *
[1] 2020/05/13 04:03:39.196727 [INF] STREAM:   Messages     :       1000000 *
[1] 2020/05/13 04:03:39.196728 [INF] STREAM:   Bytes        :     976.56 MB *
[1] 2020/05/13 04:03:39.196730 [INF] STREAM:   Age          :     unlimited *
[1] 2020/05/13 04:03:39.196731 [INF] STREAM:   Inactivity   :     unlimited *
[1] 2020/05/13 04:03:39.196732 [INF] STREAM: ----------------------------------
[...]

PR #59, opened 2020-05-17
helm: stan, volumeClaim configuration

mcuadros:
This allows to use the volumeClaimTemplate for the logs and the store. Otherwise, the cluster cannot recover their state, see https://github.com/nats-io/nats-streaming-server/issues/547

https://github.com/nats-io/nats-streaming-server/pull/548/files

Also, the support for choosing the storageClass has been introducing, and the default size has been changed to something more performance since usually small disk have low I/O in cloud platforms.

### DIFF
--- a/helm/charts/stan/values.yaml
+++ b/helm/charts/stan/values.yaml
@@ -61,12 +61,21 @@ store:
   # 
   limits: {}
 
+  #
+  # Volume claim configuration. Required if the store type is `file` or if
+  # cluster it's enabled.
+  #
+  volume:
+    enabled: true
+    mount: /data/stan
+    storageSize: 100Gi
+    storageClass:
+
   # 
   # File storage settings.
   # 
   file:
     path: /data/stan/store
-    storageSize: 1Gi
 
   # In case of using file or sql storage, sets up a 3 node cluster.
   cluster:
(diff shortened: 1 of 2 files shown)


wallyqs:
Thanks @orefalo I tried locally didn't get an error but maybe have introduced an issue during an upgrade, I will take another look
```
helm install stan ./helm/charts/stan 
```

samstride:
@wallyqs , Hi, just checking timelines for pushing this chart into helm stable repo? It looks like it will fix #58  for me. Cheers.

wallyqs:
Hey @samstride sorry for the wait, I'm aiming at releasing a new version the charts in a couple of days.

orefalo:
Thanks, will try again in a few days (with the new chart) and provide the complete log.

helm/charts/stan/values.yaml
```diff
@@ -61,12 +61,21 @@ store:
...
   limits: {}
 
+  #
+  # Volume claim configuration. Required if the store type is `file` or if
+  # cluster it's enabled.
+  #
+  volume:
+    enabled: true
+    mount: /data/stan
+    storageSize: 100Gi
```
ColinSullivan1, line 71: 1Gi -> 100Gi.  Sounds good...   Were there any issues?

wallyqs: I think I will lower this down back to 1Gi by default otherwise will get an error when upgrading since cannot be changed on the fly.

PR #60, opened 2020-05-25
[helm] Add access mode and replicas for FT mode

wallyqs:
It should be possible to set in ReadWriteMany mode to support FT groups using volumes.

Example:

```yaml
stan:
  replicas: 2
  nats:
    url: "nats://my-nats:4222"

store:
  type: file

  # 
  # Fault tolerance group
  # 
  ft:
    group: foo

  # 
  # File storage settings.
  # 
  file:
    path: /data/stan/store

  # volume for EFS
  volume:
    mount: /data/stan
    storageSize: 1Gi
    storageClass: aws-efs
    accessModes: ReadWriteMany
```

### DIFF
--- a/helm/charts/stan/README.md
+++ b/helm/charts/stan/README.md
@@ -47,6 +47,21 @@ stan:
       natsClusterName: my-nats # Name of NATS cluster created by NATS Operator
 ```
 
+### Number of replicas
+
+In case of using fault tolerance mode, you can set the number of replicas
+to be usef in the FT group.
+
+```
+stan:
+  replicas: 2
+
+# Note: in case of using clustering you will always get 3 replicas.
+store:
+  cluster:
+    enabled: true
+```
+
 ### Server Image
 
 ```yaml
@@ -98,6 +113,11 @@ In case of using a shared volume that supports a `readwritemany`,
 you can enable fault tolerance as follows.
 
 ```yaml
+stan:
+  replicas: 2
+  nats:
+    url: "nats://my-nats:4222"
+
 store:
   type: file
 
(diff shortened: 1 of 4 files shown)


helm/charts/stan/README.md
```diff
@@ -47,6 +47,21 @@ stan:
       natsClusterName: my-nats # Name of NATS cluster created by NATS Operator
 ```
 
+### Number of replicas
+
+In case of using fault tolerance mode, you can set the number of replicas
+to be usef in the FT group.
```
ColinSullivan1, line 53: Typo, `used`.

helm/charts/stan/README.md
```diff
@@ -47,6 +47,21 @@ stan:
...
+
+```
+stan:
+  replicas: 2
+
+# Note: in case of using clustering you will always get 3 replicas.
```
ColinSullivan1, line 59: Exactly 3 or at least 3?  e.g. if I set 5 replicas, will that work?

helm/charts/stan/README.md
```diff
@@ -98,21 +113,33 @@ In case of using a shared volume that supports a `readwritemany`,
...
+
 store:
   type: file
 
   # 
   # Fault tolerance group
   # 
   ft:
-    group: 
+    group: foo
```
ColinSullivan1, line 128: nitpick...   maybe name it something like `my-group` to be consistent with the `my-nats` service

helm/charts/stan/README.md
wallyqs: with cluster enabled you will get 3 replicas only, even if the number of replicas is changed.

PR #62, opened 2020-06-07
Allow override name of resources

malvex:
Hi, I was trying to use helm charts as requirement for my chart like this:

```
dependencies:
  - name: nats
    repository: https://nats-io.github.io/k8s/helm/charts/
    version: 0.4.0
  - name: stan
    repository: https://nats-io.github.io/k8s/helm/charts/
    version: 0.4.0
```
However because `.Release.Name` is used as a name for all resources it fails with "resource already exists" error because both nats and stan chart is trying to create resource `MY-RELEASE-config` and the rest.

This change doesn't break anything and allows to override name of each chart with values.yaml. It is widely used approach.

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - nats
   - messaging
   - cncf
-version: 0.4.0
+version: 0.4.1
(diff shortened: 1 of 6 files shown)


palsivertsen:
I suspect this broke all the charts over at https://hub.helm.sh/charts/nats

PR #77, opened 2020-06-23
Add prometheus annotations if exporter is enabled.

### LINKED ISSUE #63: helm: add support for Prometheus service discovery
The default Prometheus service discovery configuration does not discover the endpoints at port 7777 exposed by prometheus-nats-exporter. One can add the endpoint manually to the Prometheus instance. Adding the Prometheus annotations enabling the scraping by default to the helm charts of stan and nats would remove this step.

E.g.
```yaml
"prometheus.io/scrape": "true"
"prometheus.io/path": "/metrics"
"prometheus.io/port": "7777"
```

Is it possible to add this feature?

bw512:
Resolves #63.

### DIFF
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -18,8 +18,13 @@ spec:
   serviceName: {{ template "nats.name" . }}
   template:
     metadata:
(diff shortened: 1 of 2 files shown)


bw512:
Wow, thanks for the quick merge!

PR #84, opened 2020-07-16
Improve Helm Chart for NATS Streaming

emaincourt:
Hi,

Thanks a lot for providing charts for both NATS and NATS Streaming ! When using them I noticed some areas of improvement. Please let me know if something looks wrong to you, or you don't think those changes are relevant at all.

**Version upgrade**

* Bump up the version of NATS Streaming to 0.18.0
* Bump up the patch version of the metrics exporter 0.6.2

**Pod annotations**

As recently merged into the NATS chart in https://github.com/nats-io/k8s/pull/72, add the ability to annotate pods. In my case very useful for Istio.

**Clustering**

Nowadays there is no opportunity to deploy a StatefulSet of more than 3 nodes. I don't think there is a reason for that but please let me know if I'm wrong. This PR allows to explicitely define the cluster size, with a validation error if less than 3.

**Bug fix**

When templating `stan.clusterPeers`, `.Release.Name` is being used. However when defining `nameOverride`, `stan.clusterPeers` should make use of it as well. Using `stan.name` instead of `.Release.Name` handles both the cases.

Thanks in advance for reading

### DIFF
--- a/helm/charts/stan/templates/_helpers.tpl
+++ b/helm/charts/stan/templates/_helpers.tpl
@@ -9,7 +9,15 @@ Expand the name of the chart.
 Return the list of peers in a NATS Streaming cluster.
 */}}
 {{- define "stan.clusterPeers" -}}
-{{- range $i, $e := until 3 -}}
-{{- printf "'%s-%d'," $.Release.Name $i -}}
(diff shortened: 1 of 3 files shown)


wallyqs:
All great fixes, thanks a lot!

Issue #106, opened 2020-08-24
helm: Support --file_slice_archive_script in the STAN chart


wallyqs:
Make it possible to define a container image that is able to backup the rotated files by accessing the same volume mount as the nats-streaming-server pod: 

- About file slice archive script in STAN
https://docs.nats.io/nats-streaming-server/configuring/persistence/file_store#options

- NATS Streaming chart: 
https://github.com/nats-io/k8s/blob/master/helm/charts/stan/templates/statefulset.yaml#L84

wallyqs:
Fixed via #107

Issue #104, opened 2020-08-21
Backing up stan file store


drpebcak:
When using the file store configuration, how can I back up the rotated files on a regular basis?

In other charts I've seen something like this done by adding another container to the pod - it could be as simple as supporting that in the chart, with a script to upload to s3 and remove the old segments from the PVC.

PR #107, opened 2020-08-26
S3 backup

drpebcak:
This is an initial, sort of naive implementation to allow backing up the stan file store. I'm looking for some feedback on approach here. Addresses #104 and #106

### DIFF
--- a/helm/charts/stan/templates/statefulset.yaml
+++ b/helm/charts/stan/templates/statefulset.yaml
@@ -181,6 +181,10 @@ spec:
                 name: {{ .Values.store.backup.credentialsSecretName }}
                 key: awsDefaultRegion
           {{- end }}
+          {{- with .Values.store.backup.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           {{- if .Values.store.volume.enabled }}
           volumeMounts:
             - name: {{ template "stan.name" . }}-pvc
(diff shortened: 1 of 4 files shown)


wallyqs:
I think this approach LGTM, I like that it is using the environment variables via a secret.  Should maybe add optional resources for the backup container pod so that it does not steal cycles from the nats-streaming server one during the backup process? Also would be good to add a usage example in the README from the chart before merging. Thanks for the contribution!

helm/charts/stan/templates/statefulset.yaml
```diff
@@ -137,6 +143,55 @@ spec:
...
           {{- end }}
+        {{- if eq .Values.store.type "file" }}
+        {{- if .Values.store.backup.enabled }}
+        - name: backup
+          image: amazon/aws-cli
+          command: ["/bin/sh", "-c"]
+          args:
+          - |
+            while true; do
+              curl -s 127.0.0.1:8222/streaming/serverz | grep 'Leader'
```
wallyqs, line 154: Talking to @kozlovic on whether only backup the Leader, and we now think that it might be simpler and safer to backup leader and follower nodes as well into different folders using the $POD_NAME, something like:

```sh
aws s3 cp {{ .Values.store.volume.mount }}/* {{ .Values.store.backup.s3Path }}/$POD_NAME
```

kozlovic: Right. Backup each node regardless of their leadership status. This would prevent issues when leadership changes and file stores are not necessarily yet at the same stage.

drpebcak: Ok, just removed the check for `Leader`, and also added a prefix using `stan.name`/`POD_NAME` to make sure different deployments of stan could use the same bucket without collisions.

PR #109, opened 2020-08-29
More backup enhancements

drpebcak:


helm/charts/stan/templates/configmap.yaml
```diff
@@ -46,7 +46,10 @@ data:
       {{- else }}
       store: "file"
       dir: {{ .Values.store.file.path }}
-      {{- end  }}
+      {{- with .Values.store.file.options }}
+      file_options: {{ toPrettyJson . | indent 6 }}
```
wallyqs, line 50: (Just as a note) since it the backup script is defined in the command line with `--file_slice_archive_script`, in case `slice_archive_script` is defined here it would not be overriden since command line flags take precedence over the options  in the config file

https://github.com/nats-io/nats-streaming-server/blob/d7aed78cd69f40e902a7aa043ba266ee295c82b9/server/conf.go#L506

PR #118, opened 2020-09-10
wait for stan to be ready before marking the pod as ready

drpebcak:
This PR adds a readiness check to the stan container.


wallyqs (approved):
LGTM

PR #119, opened 2020-09-11
enhancements for backup functionality

drpebcak:
This PR makes the backup container script configurable, and by default it ensures that only .bak files that are present at the beginning of the loop will be deleted at the end - that way, no slices will be deleted if they got rotated halfway through the sync.


wallyqs (approved):
lgtm

PR #121, opened 2020-09-11
helm: ReadinessProbe for clustering mode

wallyqs:
This adds another readinessprobe option which can be used to wait until there is quorum each time that a pod is restarted during upgrades.  This requires using the `nats-streaming:alpine` image so currently not enabled by default and have to opt-in with `clusterReadinessProbe.enabled=true`, but this may change in the future.

/cc @drpebcak @kozlovic 

Example:

```yaml
clusterReadinessProbe:
  enabled: true

stan:
  image: nats-streaming:alpine
  replicas: 3
  nats:
    url: "nats://my-nats:4222"

store:
  type: file

  cluster:
    enabled: true

  #
  # File storage settings.
  #
  file:
    path: /data/stan/store

  # Volume for each pod.
  volume:
    enabled: true

    # Mount path for the volume.
    mount: /data/stan
```

### DIFF
--- a/helm/charts/stan/values.yaml
+++ b/helm/charts/stan/values.yaml
@@ -81,8 +81,8 @@ clusterReadinessProbe:
   enabled: false
   probe:
     exec:
-      command: ["/bin/sh", "-c", "n=$(($STAN_REPLICAS-1)); for i in `seq 0 $n`; do wget -qO-  $STAN_SERVICE_NAME-$i.$STAN_SERVICE_NAME:8222/streaming/serverz 2> /dev/null > /dev/null; done; if [[ $? -ne 0 ]]; then exit 0; fi; wget -qO- 127.0.0.1:8222/streaming/serverz | grep role | awk '{print $2}' | grep '\\(Follower\\|Leader\\)'"]
-    initialDelaySeconds: 10
+      command: ["/bin/sh", "-c", "n=$(($STAN_REPLICAS-1)); for i in `seq 0 $n`; do wget -qO- $STAN_SERVICE_NAME-$i.$STAN_SERVICE_NAME:8222/streaming/serverz 2> /dev/null > /dev/null; done; if [[ $? -ne 0 ]]; then wget -qO- 127.0.0.1:8222/streaming/serverz | grep role | awk '{print $2}' | grep '\\(Candidate\\|Follower\\|Leader\\)'; else wget -qO- 127.0.0.1:8222/streaming/serverz | grep role | awk '{print $2}' | grep '\\(Follower\\|Leader\\)'; fi;"]
+    initialDelaySeconds: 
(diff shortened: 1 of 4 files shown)


helm/charts/stan/values.yaml
```diff
@@ -73,6 +73,18 @@ readinessProbe:
...
 
+# Readiness probe to determine when nats-streaming cluster is ready.
+# Overrides the readinessProbe in case present.
+# NOTE: Only works with the nats-streaming alpine image.
+# ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+clusterReadinessProbe:
+  enabled: false
+  probe:
+    exec:
+      command: ["/bin/sh", "-c", "n=$(($STAN_REPLICAS-1)); for i in `seq 0 $n`; do wget -qO- $STAN_SERVICE_NAME-$i.$STAN_SERVICE_NAME:8222/streaming/serverz 2> /dev/null > /dev/null; done; if [[ $? -ne 0 ]]; then wget -qO- 127.0.0.1:8222/streaming/serverz | grep role | awk '{print $2}' | grep '\\(Candidate\\|Follower\\|Leader\\)'; else wget -qO- 127.0.0.1:8222/streaming/serverz | grep role | awk '{print $2}' | grep '\\(Follower\\|Leader\\)'; fi;"]
```
wallyqs, line 84: Slightly more readable version of the script here:

```sh
# Statefulset starts from index 0 so up to n-1
n=$(($STAN_REPLICAS-1)); 

# Check if there are any peers and in case any of them fails, then consider the cluster not ready
# and let the readiness probe succeed in case we are also candidate and ready.
for i in `seq 0 $n`; do 
  wget -qO- $STAN_SERVICE_NAME-$i.$STAN_SERVICE_NAME:8222/streaming/serverz 2> /dev/null > /dev/null; 
done; 

# If cluster was not ready, then check for candidate state until there are more peers,
# otherwise wait until there is always quorum in the cluster.
if [[ $? -ne 0 ]]; then 
  wget -qO- 127.0.0.1:8222/streaming/serverz | grep role | awk '{print $2}' | grep '\\(Candidate\\|Follower\\|Leader\\)'; 
else 
  wget -qO- 127.0.0.1:8222/streaming/serverz | grep role | awk '{print $2}' | grep '\\(Follower\\|Leader\\)'; 
fi;
```

drpebcak:
Looks good to me @wallyqs!

One thing I noticed is that the serviceaccount name is hardcoded so you can't deploy two stan instances in the same namespace. Here is a PR to fix that - #122 - it defaults to the old value so it should be fine for backwards compatibility.

wallyqs:
Thanks for checking @drpebcak. OK merging.

PR #122, opened 2020-09-12
make serviceaccountnames dynamic

drpebcak:


helm/charts/stan/templates/_helpers.tpl
```diff
@@ -21,3 +21,10 @@ Return the list of peers in a NATS Streaming cluster.
 {{- end -}}
 {{ print $replicas }}
 {{- end -}}
+
+{{/*
+Define the serviceaccountname
+*/}}
+{{- define "stan.serviceAccountName" -}}
+{{- default "nats-streaming" .Values.serviceAccountName | trunc 63 | trimSuffix "-" -}}
```
wallyqs, line 29: Thanks, we'll likely have to keep this default `nats-streaming` name around to be backwards compatible.

wallyqs (approved):
LGTM!

PR #124, opened 2020-09-24
Add jetstream single node example

wallyqs:
Adds a couple of examples for a simple quickstart of JetStream using the nightly images:

```sh
kubectl apply -f nats-server/simple-nats.yml 
kubectl apply -f nats-server/nats-js-leaf.yaml

kubectl port-forward nats-0 4222:4222
nats str add ORDERS --subjects "ORDERS.*" --ack --max-msgs=-1 --max-bytes=-1 --max-age=1y --storage file --retention limits --max-msg-size=-1 --discard=old
nats con add ORDERS NEW --filter ORDERS.received --ack explicit --pull --deliver all --max-deliver=-1 --sample 100
nats con add ORDERS DISPATCH --filter ORDERS.processed --ack explicit --pull --deliver all --max-deliver=-1 --sample 100
nats con add ORDERS MONITOR --filter '' --ack none --target monitor.ORDERS --deliver last --replay instant
```

### DIFF
--- a/nats-server/nats-js-leaf.yaml
+++ b/nats-server/nats-js-leaf.yaml
@@ -2,37 +2,37 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: nats-config
+  name: nats-js-config
 data:
   nats.conf: |
     pid_file: "/var/run/nats/nats.pid"
     http: 8222
+    
+    debug: true
+    trace: true
 
-    cluster {
-      port: 6222
-      routes [
-        nats://nats-0.nats.default.svc:6222
-        nats://nats-1.nats.default.svc:6222
-        nats://nats-2.nats.default.svc:6222
-      ]
-
-      cluster_advertise: $CLUSTER_ADVERTISE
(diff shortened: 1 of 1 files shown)


nats-server/nats-js-leaf.yaml
```diff
@@ -0,0 +1,161 @@
...
+  nats.conf: |
+    pid_file: "/var/run/nats/nats.pid"
+    http: 8222
+    
+    debug: true
+    trace: true
+
+    jetstream {
+      store_dir: "/data/jetstream/store"
+      max_file_store: 1G
```
variadico, line 16: Does it make sense to also set `max_memory_store`? Are users allowed to pick memory storage?

variadico (approved):

nats-server/nats-js-leaf.yaml
wallyqs: they do though we can do that for the helm charts instead and keep the example for file storage for now

PR #127, opened 2020-09-26
fix alternate serviceaccountname 

drpebcak:
also modify backup script to avoid cranking resources during backup. This also serves the purpose of mitigating issues when individual files file to back up for whatever reason.


wallyqs:
Just did a release but forgot to include this 🤦, merging...

wallyqs (approved):

PR #142, opened 2020-10-22
Add Jetstream file storage options with persistent volume to NATS server helm chart

eoneal89:
Adds support to deploy nats server with Jetstream file storage mounted to a volume. Volume can be created from a PVC or from options defined in `values.yaml`

### DIFF
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -69,7 +69,19 @@ data:
     # NATS JetStream                  #
     #                                 #
     ###################################
-    jetstream: {{ .Values.nats.jetstream.enabled }}
+    jetstream {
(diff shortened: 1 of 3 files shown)


wallyqs:
Thanks for the PR! I will take a look and make a release tomorrow with the changes.

wallyqs:
Thanks again for the contribution, I will add in a few fixes since from this PR I found out that it is needed to declare mem_store in order for the changes of file_storage to be reflected (that is getting fixed via this PR: https://github.com/nats-io/nats-server/pull/1675)
Also this made me realize that might would be nice to suppot units like `GiB`, `KiB` [in the conf format](https://github.com/nats-io/nats-server/blob/master/conf/parse.go#L251-L278) so that the declaration is more compatible with how Kubernetes handle these.

PR #144, opened 2020-10-23
Add NATS Operator chart

variadico:
This adds the NATS Operator chart, copied from the nats-operator repo.

I updated the chart API version and ran `helm lint`. Then I deployed it on a cluster and verified that I was able to connect to NATS.

### DIFF
--- a/helm/charts/nats-operator/crds/customresourcedefinition.yaml
+++ b/helm/charts/nats-operator/crds/customresourcedefinition.yaml
@@ -1,0 +1,35 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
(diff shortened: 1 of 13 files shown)


helm/charts/nats-operator/crds/customresourcedefinition.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
wallyqs, line 1: think we need to update this one for non beta (newer clusters like 1.18)?

variadico: Will do in follow-up PR because this requires adding schema validation.

PR #158, opened 2020-11-10
fix(nats helm): parse serviceMonitor labels and annotations

eoneal89:
Fixes issue #157

PR #164, opened 2020-11-16
Fix labels and annotations on STAN's ServiceMonitor

Alirun:
Following pattern from #158 
Fixing the same issue with STAN's ServiceMonitor


wallyqs (approved):
LGTM

PR #165, opened 2020-11-17
nats chart: Make terminationGracePeriodSeconds overridable

gnarea:
The default 60 seconds makes it very hard to iterate when using nats for testing purposes.


wallyqs (approved):
LGTM, thanks

PR #170, opened 2020-11-21
NATS Streaming: Fix docs to reflect min number of replicas with clustering enabled

gnarea:
If I enable clustering without setting the replicas (or without setting them to 3 or more), I get the following error:

```
.Values.stan.replicas should be greater or equal to 3 in clustered mode
```


wallyqs:
Thanks!

PR #172, opened 2020-12-02
Stan chart: Allow "max_bytes" limit to have a suffix or be long

gnarea:
The current code fails when trying to use 1MB or more, even if specified as the integer `1000000` because `| int` converts it to `1e+06`, which is invalid and the pod errors out with:

```
parameter "max_bytes" value is expected to be int64, got string
```

This change allows `max_bytes` to be something like `1MB` or `1000000`.

### DIFF
--- a/helm/charts/stan/templates/configmap.yaml
+++ b/helm/charts/stan/templates/configmap.yaml
@@ -97,8 +97,8 @@ data:
         max_msgs: {{ .max_msgs | int }}
         {{- end }}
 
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

gnarea:
Thanks @wallyqs! Would you mind doing a release when you get a chance?

wallyqs:
thanks @gnarea, ok will make a release soon

PR #173, opened 2020-12-03
NATS chart: Fix setting of resources

gnarea:
The change in d13d5034b9dca5a564c5c4c90291441f39e088ee doesn't work: It has references to the non-existing value `.Values.nats.requests` and also `.Values.nats.limits` (which exists but has a totally different purpose).

I'm making this generic and taking any Kubernetes `resources` value, which is what you tend to see in Kubernetes charts.

### DIFF
--- a/helm/charts/nats/README.md
+++ b/helm/charts/nats/README.md
@@ -512,6 +512,8 @@ nats:
       memory: 6Gi
 ```
 
+No resources are set by default.
+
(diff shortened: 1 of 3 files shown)


wallyqs:
Thanks @gnarea for fixing, will do release of the charts later today.

gnarea:
Thanks @wallyqs !

PR #174, opened 2020-12-04
NATS chart: Optionally enable service topology

gnarea:
https://kubernetes.io/docs/concepts/services-networking/service-topology/


gnarea:
You're welcome! Would you mind doing a release when you get a chance? Thanks!

wallyqs:
ok new version of chart should be available after repo updaet

PR #177, opened 2020-12-07
NATS Chart working as subchart

astraw38:
* Use 'include' instead of template
* Use a fullname that will contain chartname if the current chart is a subchart.
* Extracted TLS config to _helpers.tpl.

ex:
helm install myrelease nats/nats will create ``myrelease``

If nats is a subchart of an umbrella chart called 'myrelease', then nats will be created: ``myrelease-nats``. The template for this change was the default template Helm creates when you run ``helm create <chart>`` 

Note: if using jetstream, I needed to use synadia/nats-server:nightly
image. Not sure if that should be called out somewhere in readme/notes.

### DIFF
--- a/helm/charts/nats/templates/NOTES.txt
+++ b/helm/charts/nats/templates/NOTES.txt
@@ -15,11 +15,11 @@ in the NATS documentation website:
 NATS Box has been deployed into your cluster, you can
 now use the NATS tools within the container as follows:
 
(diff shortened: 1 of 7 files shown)


wallyqs:
Thanks for the PR, I will give it a try today.

variadico:
Migrating this change to a different PR: https://github.com/nats-io/k8s/pull/192
It's exactly the same, just rebased.

PR #187, opened 2021-01-22
js: Add default for cluster name when JetStream enabled

### LINKED ISSUE #185: nats jetstream + cluster error
Clustering does not work with jetstream

Using the helm chart to install, with only jetstream enabled it starts correctly.

```
helm install nats nats/nats --version 0.7.5 --set nats.image=synadia/nats-server:nightly-20210115 --set nats.jetstream.enabled=true --set natsbox.enabled=false
```

Clustering by itself also works correctly

```
helm install nats nats/nats --version 0.7.5 --set nats.image=synadia/nats-server:nightly-20210115 --set natsbox.enabled=false --set cluster.enabled=true 
```

But if I add in clustering and jetstream

```
helm install nats nats/nats --version 0.7.5 --set nats.image=synadia/nats-server:nightly-20210115 --set nats.jetstream.enabled=true --set natsbox.enabled=false --set cluster.enabled=true 
```

The pod crashes with this log message 

```
$ k logs nats-0 -c reloader
2021/01/15 20:44:45 Starting NATS Server Reloader v0.6.0+git
2021/01/15 20:44:45 Error: open /var/run/nats/nats.pid: no such file or directory
2021/01/15 20:44:47 Error: open /var/run/nats/nats.pid: no such file or directory
2021/01/15 20:44:49 Error: open /var/run/nats/nats.pid: no such file or directory
2021/01/15 20:44:51 Error: open /var/run/nats/nats.pid: no such file or directory
2021/01/15 20:44:53 Error: open /var/run/nats/nats.pid: no such file or directory
```

The cluster is a self hosted cluster running on Ubuntu and is running kubernetes 1.17.11.

wallyqs:
Fixes #185 

### DIFF
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -97,8 +97,16 @@ data:
     cluster {
       port: 6222
 
+      {{- if .Values.nats.jetstream.enabled }}
+      {{- if .Values.cluster.name }}
+      name: {{ .Values.cluster.name }}
+      {{- else }}
+      name: {{ template "nats.name" . }}
+      {{- end }}
+      {{- else }}
       {{- with .Values.cluster.name }}
       name: {{ . }}
+      {{- end }}
       {{- end }}
 
       {{- with .Values.cluster.tls }}
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - nats
   - messaging
   - cncf
-version: 0.7.5
+version: 0.8.0
 home: http://github.com/nats-io/k8s
 maintainers:
   - name: Waldemar Quevedo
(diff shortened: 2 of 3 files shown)


helm/charts/nats/templates/configmap.yaml
```diff
@@ -97,9 +97,17 @@ data:
...
 
+      {{- if .Values.nats.jetstream.enabled }}
+      {{- if .Values.cluster.name }}
+      name: {{ .Values.cluster.name }}
+      {{- else }}
+      name: {{ template "nats.name" . }}
+      {{- end }}
+      {{- else }}
       {{- with .Values.cluster.name }}
       name: {{ . }}
```
variadico, line 108: Does this `name: {{ . }}` also need to have the if guard from above?

```yaml
      {{- if .Values.cluster.name }}
      name: {{ .Values.cluster.name }}
      {{- else }}
      name: {{ template "nats.name" . }}
      {{- end }}
```

variadico (approved):

helm/charts/nats/templates/configmap.yaml
wallyqs: No since the `with` will do the nil check instead of the `if`.

ColinSullivan1:
LGTM.

PR #199, opened 2021-02-22
Fix nats-box deployment yaml

erebe:
Volumes section in not at the correct position, so when credential or tls is enabled, helm deployment fails with 

```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec): unknown field "volumes" in io.k8s.api.apps.v1.DeploymentSpec
```

To reproduce, enable tls of nats in values.yaml


wallyqs (approved):
LGTM, thanks

wallyqs (approved):
LGTM

PR #209, opened 2021-03-23
no_tls added in configmap

SahilNarans:


helm/charts/nats/templates/configmap.yaml
```diff
@@ -254,6 +254,8 @@ data:
         {{- end }}
         }
       {{- end }}
+      {{- else }}
+      no_tls: {{ .Values.websocket.no_tls }}
```
wallyqs, line 258: Thanks a lot for the PR! Just fyi that I will change it to camel case to be consistent with the rest of the YAML fields:
```suggestion
      no_tls: {{ .Values.websocket.noTLS }}
```

jacobbeasley:
Thank you!!!

PR #213, opened 2021-04-06
[helm/nats] add priority class name for NATS helm chart

### LINKED ISSUE #212: [helm/nats] priority class is missing in values file
When NATS is considered a critical piece of infrastructure it is important to be able to boost its priority class. As of now it is not possible in the current nats helm chart.

kuznero:
Fixes #212


wallyqs (approved):
Thanks!

PR #220, opened 2021-04-12
Explicit protocol selection

akubala:
Thanks to `appProtocol` field we are able to specify an application protocol for each Service port.
It can be helpfull in some situations, e.g. for Istio to explicitly select protocol for TLS.

This feature is in stable state for Kubernetes v1.20.

Sources:
- https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol
- https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

### DIFF
--- a/helm/charts/nats/templates/service.yaml
+++ b/helm/charts/nats/templates/service.yaml
@@ -26,13 +26,19 @@ spec:
   {{- end }}
   - name: client
     port: 4222
+    appProtocol: tcp
   - name: cluster
     port: 6222
+    appProtocol: tcp
   - name: monitor
     port: 8222
+    appProtocol: http2
   - name: metrics
     port: 7777
+    appProtocol: tcp
   - name: leafnodes
     port: 7422
+    appProtocol: tcp
   - name: gateways
     port: 7522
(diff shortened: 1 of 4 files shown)


helm/charts/nats/templates/service.yaml
```diff
@@ -26,13 +26,19 @@ spec:
   {{- end }}
   - name: client
     port: 4222
+    appProtocol: tcp
   - name: cluster
     port: 6222
+    appProtocol: tcp
   - name: monitor
     port: 8222
+    appProtocol: http2
```
wallyqs, line 35: We don't use in `http2` in the server, should we change to `http` or would you recommend labeling this as `http2`?

akubala: My bad! I have changed it to `http` :)

pysiekytel:
@wallyqs is there anything else that blocks merge of that pull request?

wallyqs:
I think there are a few entries that have http2, maybe it is ok?

wallyqs:
Fixed via https://github.com/nats-io/k8s/pull/245

PR #222, opened 2021-04-14
Support setting the securityContext

agrimmer:
The Helm chart does not allow to set the securityContext because its default value is null. 
You get the following warning by Helm `coalesce.go:160: warning: skipped value for securityContext: Not a table.`

Fix: Use an empty object.


wallyqs:
Thanks, LGTM

PR #236, opened 2021-05-17
Add support for resources in remaining containers

bruth:
The use case is simply that I have an environment which has a policy for requiring explicit resource requests and limits.


wallyqs (approved):

bruth:
Thanks @wallyqs

PR #238, opened 2021-05-20
Added recommended kubernetes labels to nats-operator Helm Chart

jonhoare:
I have added some recommended kubernetes labels to the `nats-operator` Helm Chart based on the following doc.
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

This is to make it easier if trying to target pods based on labels such as `app.kubernetes.io/name`

The labels have been added to the `_helpers.tpl` and included in the `deployment.yaml`


wallyqs (approved):
LGMT!

PR #245, opened 2021-05-26
Allow to specify appProtocol

ChristophKuper:
This is needed if Nats will be used next to Istio sidecars.
https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/


nickpoorman:
@ChristophKuper @wallyqs Applying the same affinity rules to natsbox as the rest of nats is causing all sorts of issues on node pools that are specifically sized for the nats servers. Seems like natsbox should have its own affinity configuration.

PR #249, opened 2021-06-05
Add nats cluster authorization block to helm charts

nickpoorman:
This PR adds the ability for a user to configure authorization between NATS cluster nodes via the helm charts.

cc @wallyqs


wallyqs (approved):
LGTM!

PR #250, opened 2021-06-05
Updating binding name to follow the naming standard

rayjanoka:
It looks like there was a previous change to replace these `:` characters but this one was missed. thanks!


wallyqs (approved):
LGTM

PR #258, opened 2021-06-27
Fix MQTT Configuration

1995parham:
I added the MQTT port on the stateful set and also show the MQTT banner in the configuration when it is enabled. also, I recommend having a configuration for MQTT port but I want to have this PR's change small so I didn't add that.


wallyqs (approved):
LGTM, thanks a lot

PR #261, opened 2021-07-01
helm/nats: fully qualify advertised routes with the k8s cluster-domain

mvisonneau:
This change reduces the number of DNS queries being made. At scale we
have had surprises/issues even with the CoreDNS `autopath` plugin enabled.


wallyqs (approved):
LGTM

PR #262, opened 2021-07-01
helm/nats: also leverage the securityContext values for nats-box deployment

mvisonneau:
Currently, there is no built-in way to make it work when restricted pod security policies are in place.


wallyqs (approved):
LGTM

PR #265, opened 2021-07-10
Separate nats and natsbox affinity

sergeyshaykhullin:
Natsbox should have it's own affinity to not to have race conditions on 3 node clusters

Screenshot_1


wallyqs (approved):
LGTM

Issue #247, opened 2021-06-02
setup.sh not working with newer version of k8


fabmade:
curl -sSL https://nats-io.github.io/k8s/setup.sh | sh

Error: unknown flag: --generator

see https://kubernetes.io/docs/reference/kubectl/conventions/

huan:
I run into this issue today and it seems related to https://github.com/kubernetes/kubernetes/pull/68132

eonward37:
Try using kubectl apply to create the setup container

```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: nats-setup
spec:
  restartPolicy: Never
  serviceAccountName: nats-setup
  containers:
    - name: nats-setup
      image: synadia/nats-setup:latest
      imagePullPolicy: Always
EOF
```

caleblloyd:
Closing due to age of issue; if experiencing in current versions please open a new issue.

PR #266, opened 2021-07-11
use k apply to create setup container

eonward:
see https://github.com/kubernetes/kubernetes/pull/68132

Addressing #247 

Any thoughts?


wallyqs (commented):
LGTM, thanks for the fix!

huan:
Thank you very much for the fix!

PR #268, opened 2021-07-12
Add namespace config to resources

cten:
This helps those who use "helm template" to not have to add namespace to the yaml files after the fact.


wallyqs (approved):
LGTM

PR #269, opened 2021-07-13
Update label used for load balancers

wallyqs:
In latest versions of the Helm Chart different label selectors are used now:

https://github.com/nats-io/k8s/blob/f006a95746ac213b6fd766cbaa49652e8646e0c9/helm/charts/nats/templates/_helpers.tpl#L44-L47

### DIFF
--- a/helm/charts/nats/README.md
+++ b/helm/charts/nats/README.md
@@ -240,7 +240,7 @@ metadata:
 spec:
   type: LoadBalancer
   selector:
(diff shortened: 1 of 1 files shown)


wallyqs:
Reference on these labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

variadico (approved):

Issue #271, opened 2021-07-15
Add authentication token support


sdague:
It would be nice if the simple token authentication was supported in the nats helm chart for setup to complete various authentication mechanisms supported by nats.

Happy to provide a PR to add that in.

wallyqs:
@sdague I think that sounds good, glad to accept a PR on this

PR #272, opened 2021-07-15
Allow using token authentication with nats

sdague:
One of the supported modes of authentication in nats is a global
token. This makes that possible to configure with the helm chart.

Closes: #271


wallyqs (approved):

wallyqs:
Thanks!

PR #274, opened 2021-07-20
Add initial nats-kafka helm chart

variadico:
This adds a basic Helm chart for nats-kafka.

Closes https://github.com/nats-io/nats-kafka/issues/9


helm/charts/nats-kafka/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+version: "0.1.0"
```
wallyqs, line 2: Here I put the version of the charts in the repo so would be 0.8.4
```suggestion
version: "0.8.4"
```

wallyqs (approved):
LGTM

PR #275, opened 2021-07-21
helm/nats: updated to nats 2.3.2

mvisonneau:


wallyqs:
thanks, latest helm chart already updated nats version https://github.com/nats-io/k8s/blob/main/helm/charts/nats/values.yaml#L7

PR #283, opened 2021-07-26
feat(helm/nats): optional nats.jetstream.fileStorage.storageClassName

chris13524:
This PR adds support for [DefaultStorageClass](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass) ([doc](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1)). This allows chart consumers to optionally specify a storage class name, leaving the default to be provided by the cluster environment.

This mirrors Traefik's method of passing `storageClassName` [here](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/pvc.yaml#L23) and [here](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml#L375).

### DIFF
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -473,5 +473,7 @@ spec:
         resources:
           requests:
             storage: {{ .Values.nats.jetstream.fileStorage.size }}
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM! Thanks for the fix

PR #285, opened 2021-07-30
Support Authorization Block on Leaf Node

1995parham:
Based on the leaf node configuration which is described [here](https://docs.nats.io/nats-server/configuration/leafnodes/leafnode_conf#authorization-block) I have added the Authorization block.


wallyqs (approved):
Thanks!!

PR #287, opened 2021-07-30
Add Tests for NATS Chart

1995parham:
I have added a very simple request/response test to make sure we have an up and running nats cluster. Also if you agree I can also add a Github pipeline like [this](https://github.com/helm/chart-testing-action) to run the chart and its test for making sure about the PRs.


1995parham:
by using this we can find out about the pub/sub working of the deployed chart.

wallyqs (approved):

wallyqs:
Thanks for the many contributions @1995parham. I think having a Github pipeline for testing the chart would be really great!

PR #289, opened 2021-08-12
Github Pipelines

1995parham:
I have added a pipeline based on Github Actions for testing charts and lining them.


1995parham:
@wallyqs there are some issues with the maintainers' list. based on [this](https://helm.sh/docs/topics/charts/#the-chartyaml-file) we need to change `github` to `url` or we can ignore the lint state for now and just use test.

wallyqs:
Thanks @1995parham I will make the change.

PR #290, opened 2021-08-13
re-pluralize gateway cert path in configmap

natw:
Just fixing a typo that looks to have been introduced in b29643fb. The cert path in the configmap and the volumeMount need to agree.

Currently, using chart version `0.8.6`, if you have gateway tls configured, the `nats-server` container cannot find the mounted secret and errors out.


natw:
hmm, looks like this was introduced in 0.8.0

wallyqs:
Thanks!

PR #291, opened 2021-08-16
[helm/nats] Expose nats-server profiling option

rwrz:
Allows users to enable profiling on the `value.yaml` file.


wallyqs (approved):
Thanks Rodrigo

PR #292, opened 2021-08-17
fix: Correct and Short Circuiting Issue

1995parham:
Helm templates don't use short-circuiting so for an and expression they check both conditions and this causes issues. This resolves what is discussed [here](https://github.com/nats-io/k8s/commit/6d010c8a8abbaa76c402a0f2db081acfa3f058da)


wallyqs (approved):
LGTM, thanks @1995parham 💯

Issue #294, opened 2021-08-17
Add support for NATS Jetstream Encryption Key


samuel-form3:
# Problem Statement
NATS chart there is no way to configure the Jetstream Encryption Key.

# Proposed Solution

Add an encryption section in jetstream configuration like the following:

```yaml
  jetstream:
    enabled: false

    ##########################
    #                        #
    #  Jetstream Encryption  #
    #                        #
    ##########################
    encryption:
      # Use key if you want to provide the key via Helm Values
      # key: "random_key"

      # Use a secret reference if you want to get a key from a secret
      # secret:
      #   name: "nats-jetstream-encryption"
      #   key: "key"
```

PR #295, opened 2021-08-17
Add support for Jetstream Encryption Key

samuel-form3:
This PR implements the issue #294 which aims to support setting the Jetstream Encryption Key on NATS servers.


wallyqs (approved):
LGTM

PR #296, opened 2021-08-17
remove trailing whitespace from nats.conf in ConfigMap

natw:
Depending on your configuration, the configmap can be rendered with
lines having trailing whitespace. This causes k8s to display the
configmap as just a big string containing literal "\n" strings, rather
than actual linebreaks.

This is most notably a problem when using `kubectl edit`, as it is
somewhat more difficult to edit the config when it's all on one line.

https://github.com/kubernetes/kubernetes/issues/36222 has a whole bunch
of discussion on the issue, in case you wanted to read about a bunch of
people yelling past each other.

Regardless of whether it's a bug or not, seems best to just avoid it in
the first place.

### DIFF
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -138,7 +138,7 @@ data:
 
       connect_retries: {{ .Values.nats.connectRetries }}
     }
-    {{ end }}
+    {{- end }}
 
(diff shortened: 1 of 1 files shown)


wallyqs:
Makes sense to just avoid this, thanks!

wallyqs (approved):

PR #307, opened 2021-08-23
Add support for secrets as config source

aditmeno:
This PR seeks to introduce the ability to pass a custom secret as the config source for NATS

I encountered an issue when I wanted to pass custom config (like creating users with passwords) which isn't supported by the helm configmap template presently

With this PR, users can pass the name of a secret which contains their custom config (along any extra configuration which isn't natively supported by the helm chart)

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - nats
   - messaging
   - cncf
-version: 0.8.8
+version: 0.8.9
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM, thanks

wallyqs:
Thanks again for the contribution, just FYI that I have changed this to be `customConfigSecret` instead to have fewer config knobs.

https://github.com/nats-io/k8s/commit/f3ea1b3bc0db4ebc6cafec4ae6108e66dc354b09

PR #310, opened 2021-08-26
Pass TLS values to jetstream controller args

dstreet:
This fixes an issue where the Helm chart values for TLS settings are not being passed as arguments to the Jetstream Controller, which prevents it from being able to communicate with Nats server setup with TLS.


wallyqs (approved):
LGTM

PR #315, opened 2021-09-02
Add origin option for nats websocket

masudur-rahman:
Signed-off-by: Masudur Rahman <masudjuly02@gmail.com>

`allowed_origins` is an important part for websocket.
To have the options available for user, I added the `same_origin` and `allowed_origins` field.

And also if `TLS` isn't enabled, `no_tls` field must be set `true`. That's why I added this.

### DIFF
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -356,7 +356,7 @@ data:
       {{- end }}
       same_origin: {{ .Values.websocket.sameOrigin }}
       {{- with .Values.websocket.allowedOrigins }}
-      allowed_origins: ["{{ first . }}" {{- range (rest .) }}, "{{ . }}" {{- end }}]
+      allowed_origins: {{ toRawJson . }}
       {{- end }}
     }
     {{- end }}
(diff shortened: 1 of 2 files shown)


masudur-rahman:
Hey @wallyqs, can you take a look at this PR ?

helm/charts/nats/templates/configmap.yaml
```diff
@@ -354,6 +354,10 @@ data:
       {{- else }}
       no_tls: {{ .Values.websocket.noTLS }}
       {{- end }}
+      same_origin: {{ .Values.websocket.sameOrigin }}
+      {{- with .Values.websocket.allowedOrigins }}
+      allowed_origins: ["{{ first . }}" {{- range (rest .) }}, "{{ . }}" {{- end }}]
```
wallyqs, line 359: Might be simpler to use `toRawJson` so that it becomes the array? 

```suggestion
      allowed_origins: {{ toRawJson . }}
```

wallyqs (approved):
LGTM, thanks for the contribution!

helm/charts/nats/templates/configmap.yaml
masudur-rahman: > Might be simpler to use `toRawJson` so that it becomes the array?

@wallyqs, thanks for the suggestion. I've updated it. PTAL

PR #317, opened 2021-09-08
fix: add support to define nats selector labels.

jdolce:
Allowing users to set custom selector labels will allow upgrades from chart <= 7.4 to 8.0 =>.

Helm Chart <=7.4 has the selector labels set to 
```
app: {{ template "nats.name" . }}
```

This has changed in Chart 8.0+ and makes it very difficult to upgrade from lower version. Without the ability to set the selector to what is already defined you will get an error similar to the following.

```
Error: UPGRADE FAILED: cannot patch "nats" with kind StatefulSet: StatefulSet.apps "nats" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```


wallyqs (approved):
LGTM

PR #321, opened 2021-09-14
[Gateway Configuration] Implicit Advertise Configuration

edribeirojunior:
Hello Guys, 
We're starting to use NATS and the Super Cluster feature, but we had a problem regarding the bootconfig , because bootconfig tries to check an ExternalIP from nodes and unfortunately our nodes do not have ExternalIps. To solve that we used an implicit advertise option passing a value to ConfigMap. 
Could you check if this is something that other persons also needed? 

Thanks!


wallyqs:
Thanks @edribeirojunior, LGTM

wallyqs (approved):

PR #326, opened 2021-09-20
Template all tls.secret.name values

### LINKED ISSUE #325: It is not possible to template `.tpl.secret.name` values
I would like to make it possible to configure NATS like this from an umbrella helm chart:

```yaml
    tls:
      secret:
        name: "{{ .Release.Name }}-{{ .Values.nameOverride }}-cert-tls"
```

But currently it is not possible (`v0.8.9`).

kuznero:
Fixes #325


wallyqs (approved):
LGTM! Thanks.

Issue #298, opened 2021-08-18
[nats helm] Manage pod labels with commonLabels


wigeric:
Chart version : 0.8.4

For now, the only labels added to pod via helm are selector labels.

We sometimes need to label the pods too.

changing the statefulset.yaml from
```yaml
spec:
  template:
    metadata:
      labels:
        {{- include "nats.selectorLabels" . | nindent 8 }}
```
to 
```yaml
spec:
  template:
    metadata:
      labels:
        {{- include "nats.labels" . | nindent 8 }}
```

would do the job as keeping the selectorLabels value and adding commonLabels

icelynjennings:
@1995parham can you verify my PR would solve this? https://github.com/nats-io/k8s/pull/330

icelynjennings:
@wigeric your original issue may have been fixed in a later release as I see that `statefulset.yaml` does include `{{- include "nats.labels" . | nindent 8 }}` in metadata now. `podLabels` have also been merged yesterday.

May be able to close this

PR #330, opened 2021-09-21
feature: add podLabels to nats helmchart

icelynjennings:
Allow templating podLabels on NATS statefulset and nats box.

Relevant issues: #298

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -229,9 +229,6 @@ tolerations: []
 # Annotations to add to the NATS StatefulSet
 statefulSetAnnotations: {}
 
-# Labels to add to the NATS StatefulSet
-statefulSetLabels: {}
-
 # Labels to add to the pods of the NATS StatefulSet
 statefulSetPodLabels: {}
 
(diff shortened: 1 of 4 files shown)


helm/charts/nats/values.yaml
```diff
@@ -229,6 +229,12 @@ tolerations: []
 # Annotations to add to the NATS StatefulSet
 statefulSetAnnotations: {}
 
+# Labels to add to the NATS StatefulSet
+statefulSetLabels: {}
```
1995parham, line 233: we have common labels for this but by this, you can have specific labels for stateful sets do you need this?

helm/charts/nats/values.yaml
```diff
@@ -229,6 +229,12 @@ tolerations: []
...
+
+# Labels to add to the pods of the NATS StatefulSet
+statefulSetPodLabels: {}
```
1995parham, line 236: we can use podLabel for this.

helm/charts/nats/values.yaml
icelynjennings: You're right - reverted.

helm/charts/nats/values.yaml
icelynjennings: This is to differentiate between nats statefulset pods and natsbox deployment pods

wallyqs (approved):
LGTM, let's give this a try.

PR #342, opened 2021-09-23
Bump lame duck duration, add auth timeout, fix some limits...

wallyqs:
- Limits were not being applied in some cases since were under the wrong block.
- Made it possible to bump the auth timeout
- Termination grace period (time the pod pre-stop sleeps before shutting down) is now 120 as in the default lame duck duration from the server.  This may help with memory based streams and snapshots when using JetStream.

Sample YAML:

```yaml

nats:
  image: synadia/nats-server:nightly-20210923

  limits:
    writeDeadline: 30s
    lameDuckDuration: 122s

  logging:
    debug: true
    trace: false

  jetstream:
    enabled: false

    memStorage:
      enabled: true
      size: "2Gi"

    fileStorage:
      enabled: true
      size: "8Gi"
      storageDirectory: /data/
      storageClassName: gp2

cluster:
  enabled: true
  # Can set a custom cluster name
  name: "nats"
  replicas: 3

auth:
  enabled: true

  timeout: "5s"

  systemAccount: sys

  # token: foo

  # nkeys:
  #   users:
  #    - user: asdf
  #      pass: asdf

  # resolver:
  #   type: full
  #   operator: foo
  #   systemAccount: asdf

  basic:
    accounts:
      sys:
        users:
        - user: sys
          pass: sys
      js:
        jetstream: true
        users:
        - user: foo
```

### DIFF
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -386,21 +386,31 @@ data:
     #                #
     ##################
     {{- if .Values.auth.resolver }}
+
     {{- if eq .Values.auth.resolver.type "memory" }}
     resolver: MEMORY
     include "accounts/{{ .Values.auth.resolver.configMap.key }}"
     {{- end }}
 
     {{- if eq .Values.auth.resolver.type "full" }}
-
     {{- if .Values.auth.resolver.configMap }}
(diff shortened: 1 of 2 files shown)


ColinSullivan1 (approved):
LGTM.

Issue #355, opened 2021-10-14
Helm: Support different k8s clusters on nats.clusterRoutes 


filhodanuvem:
I'm trying to create a nats cluster which includes nats nodes on different kubernetes clusters. 
Using the helm chart, I cannot explicitly set the cluster routes expected in the configMap, it generates it based on the current k8s cluster. 

https://github.com/nats-io/k8s/blob/857edb21e71e029c9a6ff2456324c6779f1bf0a0/helm/charts/nats/templates/configmap.yaml#L135-L137

https://github.com/nats-io/k8s/blob/dbda2a6a3d47f177c884eff7831dd76eb05092e1/helm/charts/nats/templates/_helpers.tpl#L67-L72

I can send a pull request to support it as well, but first of all I would like to understand if this was done in this way because of a deeper limitation or whether I misunderstood something.

filhodanuvem:
Hello @wallyqs, do we have any plans on releasing a new version for nats chart in a near future? I'm asking just to align the expectation of my team that depends on the latest changes.

PR #356, opened 2021-10-18
Support different k8s clusters on nats helm chart

filhodanuvem:
It's not possible to create a nats cluster which includes nats nodes
on different kubernetes clusters.
Using the helm chart, I cannot explicitly set the cluster routes expected
in the configMap since it generates it based on the current k8s cluster.
This commit makes it possible with a new value nats.extraRoutes and
nats.serverName.

Relates to #355

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -48,10 +48,6 @@ nats:
 
   resources: {}
 
-  # Explicitly set routes for clustering.
-  # When JetStream is enabled, the serverName must be unique in the cluster.
-  extraRoutes: []
-
   # Server settings.
   limits:
     maxConnections:
@@ -249,6 +245,11 @@ cluster:
   enabled: false
   replicas: 3
   noAdvertise: false
+
+  # Explicitly set routes for clustering.
+  # When JetStream is enabled, the serverName must be unique in the cluster.
+  extraRoutes: []
+
   # authorization:
   #   user: foo
   #   password: pwd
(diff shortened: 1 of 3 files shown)


helm/charts/nats/values.yaml
```diff
@@ -7,6 +7,9 @@ nats:
   image: nats:2.6.1-alpine
   pullPolicy: IfNotPresent
 
+  # The servers name, shows up in logging.
+  serverName: ""
```
wallyqs, line 11: We could add this though think this means that `serverName` would only be possible to be used with single node setups since it has to be unique across the cluster, this gives some control for cases with single node leafnodes maybe.

filhodanuvem: Not sure if I got your point here, sorry. I've tested the extraRoutes to generate a nats cluster where each node has jetstream + leafnode enabled, and yes, it complained that the server name of each node must be unique inside the cluster. 
If you think we could solve this in a different way, I'm more than happy to change it.

wallyqs: Thanks for the info! OK let's merge as is for now.

filhodanuvem:
Thanks @wallyqs for merging, I've noticed that the `serverName`  implementation still incompatible with some situations. 
Eg.: When we tried to create two nodes per kubernetes cluster, it would try to use the same serverName. 
I'm going to send another pull request to change to `serverNamePrefix`, does that sound ok for you?

PR #367, opened 2021-10-28
Add nack accounts

variadico:
This adds NACK Accounts to the JetStream Controller. Other half of https://github.com/nats-io/nack/pull/42


sergeyshaykhullin:
@wallyqs Why do we have multiple places with CRDs?

1. k8s repo https://github.com/nats-io/k8s/blob/main/nack/crds.yml
2. NACK repo https://github.com/nats-io/nack/blob/main/deploy/crds.yml

That is the preferred file to follow?

wallyqs:
both are the same, but good point maybe should just keep the one in the nack repo.

sergeyshaykhullin:
Yes, right now they are the same, but not in a future)

wallyqs:
ok removed from the k8s repo and should be single copy in nack now

PR #380, opened 2021-11-10
Fix pingInterval in nats values

filhodanuvem:
I've noticed that to define pingInterval was not well documented in values.yaml. It should be on `.Values.nats.limits.pingInterval`

https://github.com/nats-io/k8s/blob/876f5e75d6b9aa3c661816ea1b25742ac8dcc0a6/helm/charts/nats/templates/configmap.yaml#L342-L344


wallyqs (approved):
Thanks!

PR #382, opened 2021-12-01
Sidecar leafnode certificates

chris13524:
This PR makes it possible to:
- provide additional containers, volumes, and volume mounts to the statefulset
- provide custom leafnodes TLS configuration options
- provide extra configuration to the reloader sidecar

PR #388, opened 2021-12-08
fix(nats): invalid template on default values

chris13524:
Fix errors introduced in #382:

```
$ helm template nats helm/charts/nats
Error: YAML parse error on nats/templates/statefulset.yaml: error converting YAML to JSON: yaml: line 50: did not find expected key
```


wallyqs (approved):
Thanks 👍

PR #403, opened 2022-01-13
Fix namespace quotation bug in nats-setup.sh

jbury:
Currently the use of double quotes horribly breaks the nats-setup script.  Confirmed to work after actually testing by building the image locally.


jbury:
Note: In addition to fixing the quote issue on the actual kubectl call, I also cleaned up some of the logic around using the `$NATS_NAMESPACE` variable.  It's now arguably cleaner (and more correct).

PR #404, opened 2022-01-13
Allow client port name update so nats can be used with istio

SamirMarin:
### What
This PR allows for the update of the client container port name. Need to be able to name the 4222 port (client port) to tcp for nats to work with istio. see here: istio/istio#28623. Reason for this is because nats is a server first protocol. 

### Test
Where I work we have copied the nats chart and made this update, we needed to do this since we are using nats with istio.

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -45,8 +45,6 @@ nats:
 
   resources: {}
 
-  clientPortName: client
-
   # Server settings.
   limits:
     maxConnections:
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -45,6 +45,9 @@ nats:
 
   resources: {}
 
+  client:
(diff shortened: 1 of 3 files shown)


helm/charts/nats/values.yaml
```diff
@@ -45,6 +45,8 @@ nats:
 
   resources: {}
 
+  clientPortName: client
```
wallyqs, line 48: How about having this as the defaults instead? 

```yaml
client:
  port: 4222
  portName: "client"
```
There is some interest in being able to change the port 4222 to port 443 to allow access in some environments so this might be more flexible?

SamirMarin: That works for me :) updated.

SamirMarin: updated, I think correctly this time.

wallyqs (approved):
LGTM! Thanks @SamirMarin

PR #427, opened 2022-02-15
adding allow_non_tls to tls config

edribeirojunior:
This will give the possibility to enable the feature flag _allow_non_tls_.


wallyqs:
Thanks for the PR! I will merge it with some changes so it will look as follows: 

```yaml
nats:
  client:
    port: 4222
  tls:
    allowNonTLS: true
    secret:
      name: nats-server-tls
    ca: "ca.crt"
    cert: "tls.crt"
    key: "tls.key"
    timeout: "5s"
```

wallyqs (commented):
LGTM

PR #442, opened 2022-02-25
[nats helm] add publishNotReadyAddresses: true to headless service

### LINKED ISSUE #441: [nats helm] Add publishNotReadyAddresses: true to Headless Service
In cluster mode, NATS uses the [Stable Network ID DNS address](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id) for routes

```
routes = [
   nats://nats-0.nats.nats.svc.cluster.local:6222,
   nats://nats-1.nats.nats.svc.cluster.local:6222,
  nats://nats-2.nats.nats.svc.cluster.local:6222,
```

With the addition of the recent `/healthz` endpoint that waits for JetStream quorom/catchup, JetStream-enabled pods start in a non-ready state and stay there

While the pod is in a non-ready state, DNS remains unresolveable, meaning that JetStream can never discover other pods

The `service.spec.publishNotReadyAddresses` field specifically mentions StatefulSet service discovery as a solution:

```
$ k explain service.spec.publishNotReadyAddresses
KIND:     Service
VERSION:  v1

FIELD:    publishNotReadyAddresses <boolean>

DESCRIPTION:
     publishNotReadyAddresses indicates that any agent which deals with
     endpoints for this Service should disregard any indications of
     ready/not-ready. The primary use case for setting this field is for a
     StatefulSet's Headless Service to propagate SRV DNS records for its Pods
     for the purpose of peer discovery. The Kubernetes controllers that generate
     Endpoints and EndpointSlice resources for Services interpret this to mean
     that all endpoints are considered "ready" even if the Pods themselves are
     not. Agents which consume only Kubernetes generated endpoints through the
     Endpoints or EndpointSlice resources can safely assume this behavior.
```

The downside to this is that Client DNS lookups will also be able to discover non-ready pods.

My recommendation is to open another issue to create a client-specefic service that will be used for Client lookups, and optimize the existing headless service for Service Discovery.

caleblloyd:
Resolves #441

### DIFF
--- a/helm/charts/nats/templates/service.yaml
+++ b/helm/charts/nats/templates/service.yaml
@@ -16,6 +16,7 @@ spec:
   selector:
     {{- include "nats.selectorLabels" . | nindent 4 }}
   clusterIP: None
+  publishNotReadyAddresses: true
   {{- if .Values.topologyKeys }}
   topologyKeys:
      {{- .Values.topologyKeys | toYaml | nindent 4 }}


wallyqs (approved):
LGTM

helm/charts/nats/templates/service.yaml
```diff
@@ -16,6 +16,7 @@ spec:
   selector:
     {{- include "nats.selectorLabels" . | nindent 4 }}
   clusterIP: None
+  publishNotReadyAddresses: true
```
wallyqs, line 19: I think it is ok to just leave it as the default instead of customizable for now, a client that is connecting externally would be using a different service such as a NodePort or LoadBalancer anyway and for internal clients we would recommend adding a different service for clients.

wallyqs (approved):
LGTM

PR #445, opened 2022-02-28
[nats helm] improve healthz detection

caleblloyd:
- switch version detection to use `regexFind`, since `semver` raises an error if no valid semver is found, and image tags are not required to be semvers
- default version to support healthz if a semver cannot be detected
- add `nats.healthcheck.detectHealthz` bool to optionally disable version detection
- remove helm upgrade checking for health check - this didn't work in some tools such as ArgoCD and should no longer be needed since #442 was merged


wallyqs (approved):
LGTM

PR #450, opened 2022-03-01
use helm-releaser

caleblloyd:
- use [helm/chart-releaser-action](https://github.com/helm/chart-releaser-action) to release chart
- git tags will change to have a release prefix such as `nats-0.13.1` and `stan-0.13.1` as opposed to `v0.13.1` with both assets
  -  allows for charts to be versioned independently
- releases will be created automatically when the `Chart.yaml` contains a new release version and is checked into `main`
- release notes will still need to be edited on the release after it is created


wallyqs (approved):
LGTM!

PR #460, opened 2022-03-03
[nats helm] faster rollouts for graceful shutdowns

caleblloyd:
Adds defaults and recommendations for total graceful shutdown with [lame duck mode](https://docs.nats.io/running-a-nats-service/nats_admin/lame_duck_mode) using the formula:

```
lame_duck_grace_period + lame_duck_duration + 20s shutdown overhead
```

~Also adds `sleep {{ .Values.nats.terminationGracePeriodSeconds }}` to end of the `preStop` lifecycle hook to avoid process receiving TERM signal during lame duck shutdown~

### DIFF
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -503,14 +503,12 @@ spec:
         lifecycle:
           preStop:
             exec:
-              # Using the alpine based NATS image, we add an extra sleep that is
-              # the same amount as the terminationGracePeriodSeconds to allow
-              # the NATS Server to gracefully terminate the client connections.
+              # send the lame duck shutdown signal to trigger a graceful shutdown
+              # nats-server will ignore the TERM signal it receives after this
               #
               command:
-              - "/bin/sh"
-              - "-c"
-              - "nats-server -sl=ldm=/var/run/nats/nats.pid"
+              - "nats-server"
+              - "-sl=ldm=/var/run/nats/nats.pid"
 
       #################################
       #                               #
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -45,7 +45,7 @@ nats:
       #  https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#probe-level-terminationgraceperiodseconds
       #  https://github.com/kubernetes/kubernetes/issues/64715
       #
-      periodSeconds: 60
+      periodSeconds: 30
       successThreshold: 1
       failureThreshold: 3
       # Only for Kubernetes +1.22 that have pod level probes enabled.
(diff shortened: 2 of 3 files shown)


helm/charts/nats/templates/statefulset.yaml
```diff
@@ -503,14 +503,13 @@ spec:
...
-              # the same amount as the terminationGracePeriodSeconds to allow
-              # the NATS Server to gracefully terminate the client connections.
+              # Using the alpine based NATS image, we send the lame duck mode graceful shutdown signal
+              # then sleep for terminationGracePeriodSeconds in order to avoid the TERM signal
               #
               command:
               - "/bin/sh"
               - "-c"
-              - "nats-server -sl=ldm=/var/run/nats/nats.pid"
+              - "nats-server -sl=ldm=/var/run/nats/nats.pid && sleep {{ .Values.nats.terminationGracePeriodSeconds }}"
```
wallyqs, line 512: I recently removed the `sleep` here to avoid making the pod hang when it exits due to the liveness check failing during the graceful shutdown which may take rollouts take longer than needed.
I think a side effect from this is that pods could take 60s to shutdown due to this k8s issue: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2238-liveness-probe-grace-period#summary

Also once the server gets the USR2 signal it will ignore the TERM that is sent soon after the preStop command finishes running: https://github.com/nats-io/nats-server/blob/444898909146d0989e64a225e4acae14ca008a20/server/signal.go#L54-L63

helm/charts/nats/values.yaml
```diff
@@ -122,12 +122,15 @@ nats:
...
-    # NOTE: this should be at least the same as 'terminationGracePeriodSeconds'
-    lameDuckDuration: "120s"
+    # grace period after pod begins shutdown before starting to close client connections
+    lameDuckGracePeriod: "10s"
 
-  # terminationGracePeriodSeconds determines how long to allow for a pod
-  # to be restarted.
-  terminationGracePeriodSeconds: 120
+    # duration over which to slowly close close client connections after lameDuckGracePeriod has passed
+    lameDuckDuration: "30s"
```
wallyqs, line 129: I think might be ok to shorten for faster rollouts by default, original 120s comes from the default of the server being 2min: https://github.com/nats-io/nats-server/blob/1712ee3707e9b252a197946fc20f3690735e9407/server/const.go#L156

helm/charts/nats/values.yaml
```diff
@@ -45,7 +45,7 @@ nats:
       #  https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#probe-level-terminationgraceperiodseconds
       #  https://github.com/kubernetes/kubernetes/issues/64715
       #
-      periodSeconds: 60
+      periodSeconds: 30
```
wallyqs, line 48: In most cases the graceful shutdown would take less than `lame duck duration` so as long as the liveness check does not happen before the pod exits, we would avoid the issue of empty pod without nats-server being idle upto termination grace period seconds.

wallyqs (approved):
LGTM, I think sounds good to have shorter defaults overall for lame duck. Not sure that the sleep on exit is still needed but shortening the rollout to take at most 60s instead of 2minutes based on the default of the server could be a good change to have faster rollouts.

helm/charts/nats/templates/statefulset.yaml
caleblloyd: oh, so then there is nothing to "fix" here in that case

helm/charts/nats/values.yaml
caleblloyd: I think that most users will probably want faster rollouts.  The [runtime configuration]](https://docs.nats.io/running-a-nats-service/configuration#runtime-configuration) states that 30s is the minimum, which is why I picked that number

helm/charts/nats/templates/statefulset.yaml
caleblloyd: I'll back out this change and document that the TERM signal will be ignored

wallyqs (approved):
LGTM, we could revise this again once the probe level termination grace period is out of beta which would improve behavior of graceful shutdown too: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#probe-level-terminationgraceperiodseconds

PR #464, opened 2022-03-08
[nats helm] use config checksum in addition to reloader

### LINKED ISSUE #443: [nats helm] Config Reloader sidecar vs Helm Managed Config
**Pros of Config Reloader sidecar:**

- Allows for external processes to modify ConfigMaps / Config Files and have configuration reloaded

*Cons of Config Reloader sidecar:*

- All pods pick up new configuration and send reload signal at the same time - potential disruption?
- Unchecked rollout of a bad configuration:

*Current Case 1*
Config: Bad but Valid
Config Reloader: Enabled
1. Helm deployment starts
2. New ConfigMap is deployed, replacing the old ConfigMap
3. The Helm deployment will succeed
4. The Config Reloader container will reload the configuration
5. Healthchecks could begin to fail, and all Pods could be marked as Not Ready

*Current Case 2*
Config: Invalid
Config Reloader: Enabled
1. Helm deployment starts
2. New ConfigMap is deployed, replacing the old ConfigMap
3. The Helm deployment will succeed
4. The Config Reloader container will attempt to reload the configuration, but NATS will reject it as invalid
5. Some time later the pod could get rescheduled, and would fail to start back up due to the invalid config

**Alternative: Helm Managed Config**

Use an annotation to automatically roll out configuration changes:
https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

This will cause StatefulSets to restart pods 1-at-a-time.  If there is a configuration-related startup issue, Helm will fail if it is run with the `--wait` flag and can optionally roll back with the `--atomic` flag

caleblloyd:
- Resolves #443 
- The Configuration file remains unchanged, but is moved out of `templates` and into `files` sub-directory so that it can be accessed by Helm `.Files` helper


helm/charts/nats/values.yaml
```diff
@@ -76,9 +76,17 @@ nats:
...
+  # Adds a hash of the ConfigMap as a pod annotation
+  # This will cause the StatefulSet to roll when the ConfigMap is updated
+  configChecksumAnnotation: true
+
   # securityContext for the nats container
   securityContext: {}
 
+  # Toggle to enable sharing process namespace in the Pod
+  # Useful for passing signals to NATS server
+  shareProcessNamespace: false
```
wallyqs, line 88: maybe also add a note to the readme under misc that this is needed to re-enable? https://github.com/nats-io/k8s/blob/main/helm/charts/nats/README.md#configuration-reload-sidecar

wallyqs (approved):
LGTM

helm/charts/nats/values.yaml
caleblloyd: The conditional to actually enable `shareProcessNamespace` on the pod does also consider the reloader - `{{- if or .Values.nats.shareProcessNamespace .Values.reloader.enabled }}`...   I can update the comment to mention that

wallyqs: ah ok, I see now. I think it is fine as is then

caleblloyd:
@wallyqs I added a section about the Config Checksum to the README

I also updated the Config Reloader documentation - and stated that it is Deprecated.  Do you agree that the Config Reloader should be deprecated and potentially removed in a further release?

wallyqs (approved):
LGTM

PR #474, opened 2022-03-24
[nats helm] Fix tests, and allow to override leafnodes and gateway ports

ashumkin:
Sometime you may want to setup several clusters on a single minikube k8s cluster run in `--driver=none` mode. Then hardcoded default ports will prevent a second cluster to be up.
Fix this with allow to override leafnodes and gateway ports.

Also, fix tests which use the default `4222` port but not the `nats.client.port` from values.


wallyqs (approved):
LGTM, thanks for the contribution

caleblloyd (approved):

PR #476, opened 2022-03-24
[surveyor helm] make TLS CA optional

caleblloyd:
Make TLS CA optional, as CA checking should fall back to the OS CA Store


wallyqs (approved):
LGTM

variadico (approved):

PR #493, opened 2022-04-20
Add Server Tags and JetStream Unique Tag

ramonberrutti-f3:
- Add server_tags option to Helm Chart.
- Add the new unique_tag option to Helm Chart.

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -10,6 +10,11 @@ nats:
   # The servers name prefix, must be used for example when we want a NATS cluster
   # spanning multiple Kubernetes clusters.
   serverNamePrefix: ""
+  
+  # Server Tags 
+  serverTags:
+  # - "for"
+  # - "bar"
 
   # Toggle profiling.
   # This enables nats-server pprof (profiling) port, so you can see goroutines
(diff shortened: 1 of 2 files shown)


helm/charts/nats/values.yaml
```diff
@@ -185,6 +185,9 @@ nats:
     # Jetstream Domain
     domain:
 
+    # Jetstream Unique Tag prevent placing a stream in the same availability zone twice.
+    unique_tag:
```
caleblloyd, line 189: the `values.yaml` file uses `camelCase` for everything, could you make this `uniqueTag`?

caleblloyd (changes requested):

helm/charts/nats/values.yaml
ramonberrutti-f3: great point. Fixed :)

ramonberrutti-f3:
@caleblloyd added also the server_tags option

helm/charts/nats/values.yaml
```diff
@@ -10,6 +10,11 @@ nats:
   # The servers name prefix, must be used for example when we want a NATS cluster
   # spanning multiple Kubernetes clusters.
   serverNamePrefix: ""
+  
+  # Server Tags 
+  tags:
```
caleblloyd, line 15: please call this `serverTags`, would like it to reflect the server configuration key as closely as possible so that users can correlate it to NATS configuration documentation easily

caleblloyd (changes requested):

caleblloyd (approved):

PR #498, opened 2022-04-27
Add warn about conflicting names in super cluster

wallyqs:
Related: https://github.com/nats-io/nats-server/issues/3082


caleblloyd (approved):

caleblloyd:
@wallyqs we could also make the default `server_name` be `{{ cluster_name }}-{{ deployment_name }}`

PR #499, opened 2022-05-04
add a cache volume for controller

cortopy:
Current nack chart allows to set `securityContext` but this doesn't work.

As per https://github.com/nats-io/nack/issues/48, the nack js controller will create a tmp directory in "/" which is not allowed if the container is not running as root

This PR adds a volume, so that the following works

```
securityContext:
  fsGroup: 1000
  runAsUser: 1000
  runAsNonRoot: true
```


variadico (approved):
Great! I was able to deploy this on a local cluster. No more CrashloopBackOff!

PR #508, opened 2022-05-24
Add index html for helm charts page

wallyqs:
This is so that when visiting https://nats-io.github.io/k8s/helm/charts/  with the browser instead of 404 we redirect to main site of repo.


caleblloyd (approved):
LGTM

Issue #510, opened 2022-06-08
Labels in request-reply test pod in nats chart clash with service selector


domg123:
When the nats chart is generated and run we noticed the service would resolve to 1 more DNS addresses than there were running nats replicas.  The extra DNS address would not accept traffic on port 4222, though the others did.

This IP address was tracked down to being the nats-test-request-reply that runs on chart installation.  Even though it has run to completion, the headless service includes it as matching DNS for the main nats service address, as the selectors match.

The request-reply test selector is ..
    {{- include "nats.labels" . | nindent 4 }}

The service selector is ...
    {{- include "nats.selectorLabels" . | nindent 4 }}

The nats.labels includes selectorLabels, so however you set selectorLabels, the service selector will always include the test pod.

This doesn't usually cause a permenant error, I'm assuming the client will try other DNS addresses, until one works -  it does seem to cause some instability for some of our local dev environments (the nats server will become inaccessible).  Fixing this, by removing the completed test pod resolves the problem, so whilst it should cope with the incorrect host in the DNS, it is the proximate cause.

The fix should be to set some different set of labels for the test (like the nats-box)

domg123:
I note - if I don't get any distractions - I'm planning to submit a fix for this shortly.

PR #511, opened 2022-06-08
Fixed helm nats chart test labels so they don't match service selector

domg123:
- Removes the use of the nats.labels from the nats chart test, so the test pod is not picked up by the service selector
- Adds in a couple of replacement labels.


domg123:
fixes #510

wallyqs (approved):
LGTM, thanks

PR #523, opened 2022-07-03
fix nats-box credential env variable

jdartigalongue:
The current environment variables set in the Helm chart are not used by the CLI. 
This PR fixes the environment variables.


caleblloyd (approved):
LGTM

PR #528, opened 2022-07-21
[nats helm] add automountServiceAccountToken

### LINKED ISSUE #527: Feature request: expose `automountServiceAccountToken` in PodSpec
Hi,

Would there please be a possibility to expose `automountServiceAccountToken` (https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) as a configurable value? Is there a reason why only `serviceAccountName` is exposed in PodSpec?

Thank you.

caleblloyd:
- Fixes #527

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -8,7 +8,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 0.17.1
+version: 0.17.2
 home: http://github.com/nats-io/k8s
 maintainers:
 - name: Waldemar Quevedo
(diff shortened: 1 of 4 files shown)


wallyqs (approved):

helm/charts/nack/Chart.yaml
```diff
@@ -7,7 +7,7 @@ keywords:
 - nats
 - jetstream
 - cncf
-version: 0.14.2
```
thisthat, line 10: I might miss something but Is this the correct version to bump? I would expect a version bump here: https://github.com/nats-io/k8s/blob/main/helm/charts/nats/Chart.yaml#L11

caleblloyd: good catch thanks

PR #531, opened 2022-07-25
[nats helm] use hasKey for automountServiceAccountToken detection

### LINKED ISSUE #529: nats helm release 0.17.2 installation fails when using default flags
running `helm install my-nats nats/nats` with version 0.17.2 leads to the following error

```
Error: INSTALLATION FAILED: template: nats/templates/statefulset.yaml:200:17: executing "nats/templates/statefulset.yaml" at <eq .Values.nats.automountServiceAccountToken false>: error calling eq: incompatible types for comparison
``` 

current workaround is to either use "--version 0.17.1" or add extra flags "--set nats.automountServiceAccountToken=true --set natsbox.automountServiceAccountToken=true"

should install work without specifying any additional flags?

### LINKED ISSUE #530: bug: `automountServiceAccountToken` defaults to null
Hello, just applied version 0.17.2 and there seems to be a small issue with the if statement:

$ helm template . --debug
install.go:172: [debug] Original chart version: ""
install.go:189: [debug] CHART PATH: /home/odubaj/KEPTN/keptn/installer/manifests/keptn

Error: template: keptn/charts/nats/templates/statefulset.yaml:200:17: executing "keptn/charts/nats/templates/statefulset.yaml" at <eq .Values.nats.automountServiceAccountToken false>: error calling eq: incompatible types for comparison
helm.go:81: [debug] template: keptn/charts/nats/templates/statefulset.yaml:200:17: executing "keptn/charts/nats/templates/statefulset.yaml" at <eq .Values.nats.automountServiceAccountToken false>: error calling eq: incompatible types for comparison

Seems like a default value of `null` should not be used in the `values.yaml` as `null` cannot be compared with boolean value, but I am not an expert in charts and it might be something else.

Thank you!

caleblloyd:
- Fixes #530 
- Fixes #529 

I cannot recreate the issue, and CI is passing, but multiple users are reporting a problem with comparing `boolean` to `nil` from #527  so this PR switches to use `hasKey` instead

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -8,7 +8,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 0.17.2
+version: 0.17.3
 home: http://github.com/nats-io/k8s
 maintainers:
 - name: Waldemar Quevedo
--- a/helm/charts/nats/templates/nats-box.yaml
+++ b/helm/charts/nats/templates/nats-box.yaml
@@ -61,7 +61,7 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- if or (eq .Values.natsbox.automountServiceAccountToken false) (eq .Values.natsbox.automountServiceAccountToken true) }}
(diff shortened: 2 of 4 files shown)


wallyqs (approved):
LGTM

PR #536, opened 2022-07-26
[nats helm] allow verify_cert_and_check_known_urls in tls config (#535)

### LINKED ISSUE #534: [nats helm] allow_non_tls should be allowNonTLS
Given this

https://github.com/nats-io/k8s/blob/7e658c542e133dcc7386903ef5aac9d320b1c93f/helm/charts/nats/templates/configmap.yaml#L56-L58

The value should be `allowNonTLS`

https://github.com/nats-io/k8s/blob/7e658c542e133dcc7386903ef5aac9d320b1c93f/helm/charts/nats/values.yaml#L257

or perhaps keep it the same as in the nats config, and update the ConfigMap template.

### LINKED ISSUE #535: [nats helm] verify_cert_and_check_known_urls option isn't configurable for Custer TLS
To be able to use TLS client authentication, the following options need configuring:

```
cluster {
  verify: true
  verify_and_map: true
  verify_cert_and_check_known_urls: true
}
```

The first two options are overridable by Helm (`verify` and `verifyAndMap`), however `verify_cert_and_check_known_urls` is not.

It looks like a helper template is filtering a specific set of options:

https://github.com/nats-io/k8s/blob/7e658c542e133dcc7386903ef5aac9d320b1c93f/helm/charts/nats/templates/_helpers.tpl#L104-L134

NickLarsenNZ:
Enable `verifyCertAndCheckKnownUrls` (valid under `cluster.tls`) by the `nats.tlsConfig` helper

- closes #535
- closes #534

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -8,7 +8,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 0.17.2
+version: 0.17.3
 home: http://github.com/nats-io/k8s
 maintainers:
 - name: Waldemar Quevedo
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM, thanks

PR #537, opened 2022-07-27
surveyor - jetstream auth for credentials

rossmerr:


caleblloyd:
I think that it may be more appropriate to reference secrets, with a value in the format:

```
accounts:
- name: user1
  credentials:
    secret:
      name: user-creds
      key: user1.creds
- name: user2
  credentials:
    secret:
      name: user-creds
      key: user2.creds
```

Then we mount the secrets and reference the keys in the config, similar to how the TLS certs work

caleblloyd:
Closing as stale

PR #539, opened 2022-07-27
[nats helm] allow default permissions to be set (#538)

### LINKED ISSUE #538: [nats helm] there doesn't appear to be a way to restrict topic access
In my docker-compose PoC, I was able to configure `default_permissions` under the `authorization` section. For example:

```
authorization: {
    default_permissions: {
      publish: ["SANDBOX.*"],
      subscribe: ["PUBLIC.>"]
    }
}
```

Without this block, any authenticated user can publishers/subscribers to any topic.

NickLarsenNZ:
- closes #538

### DIFF
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -537,17 +537,38 @@ data:
     no_auth_user: {{ . }}
     {{- end }}
 
-    {{- with .users }}
+    {{- if or .users (or .timeout .defaultPermissions) }}
     authorization {
       {{- if $.Values.auth.timeout }}
       timeout: {{ $.Values.auth.timeout }}
       {{- end }}
 
+      {{- with .users }}
       users: [
         {{- range . }}
         {{- toRawJson . | nindent 8 }},
         {{- end }}
       ]
+      {{- end }}
+
+      {{- with $.Values.auth.basic.defaultPermissions }}
+      default_permissions: {
+        {{- if .publish }}
+        publish: [
+          {{- range .publish }}
+          {{- toRawJson . | nindent 10 }},
+          {{- end }}
+        ],
+        {{- end }}
+        {{- if .subscribe }}
+        subscribe: [
+          {{- range .subscribe }}
+          {{- toRawJson . | nindent 10 }},
+          {{- end }}
+        ],
+        {{- end }}
+      }
+      {{- end }}
     }
     {{- end }}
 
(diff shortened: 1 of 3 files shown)


helm/charts/nats/templates/configmap.yaml
```diff
@@ -537,17 +537,38 @@ data:
     no_auth_user: {{ . }}
     {{- end }}
 
-    {{- with .users }}
+    {{- if or .users (or .timeout .defaultPermissions) }}
```
NickLarsenNZ, line 540: There's probably another way to check if `auth.basic` has items inside it.

helm/charts/nats/templates/configmap.yaml
```diff
@@ -537,17 +537,38 @@ data:
...
+      {{- end }}
+
+      {{- with $.Values.auth.basic.defaultPermissions }}
+      default_permissions: {
+        {{- if .publish }}
+        publish: [
+         {{- range .publish }}
+         {{- toRawJson . | nindent 10 }},
+         {{- end }}
+        ],
```
NickLarsenNZ, line 561: trailing commas seem to be ok (I ran it on my cluster and saw no errors).

JustinCrawshayCymonz (approved):

NickLarsenNZ:
@wallyqs just wondering if you're ok with this?

wallyqs:
Thanks for the PR @NickLarsenNZ, I think looks good overall, will try to merge soon.

helm/charts/nats/templates/configmap.yaml
```diff
@@ -537,17 +537,38 @@ data:
...
+      {{- with .users }}
       users: [
         {{- range . }}
         {{- toRawJson . | nindent 8 }},
         {{- end }}
       ]
+      {{- end }}
+
+      {{- with $.Values.auth.basic.defaultPermissions }}
+      default_permissions: {
```
wallyqs, line 555: One thing to note is that this will work for when not using accounts:

```
auth:
  enabled: true
  basic:
    defaultPermissions:
      publish: ["hello.>"]
      subscribe: ["hello.>"]
    users:
      - user: one
        pass: one
```

In case of using accounts, then have to use snake_case right now since we only apply using toRawJSON

```yaml
auth:
  enabled: true
  accounts:
      sys:
        users:
        - user: sys
          pass: sys
      js:
        jetstream: true
        default_permissions:
          publish: ["js.hello.>"]
          subscribe: ["js.hello.>"]
        users:
        - user: foo
          pass: foo
        - user: bar
          pass: bar
      notjs:
        users:
        - user: quux
          pass: quux
```

wallyqs (approved):
LGTM

PR #542, opened 2022-07-29
add insecure variables

vankosa:
https://docs.nats.io/running-a-nats-service/configuration/leafnodes/leafnode_conf#tls-configuration-block

An important parameter that is often needed when using external access to LeafNodes

### DIFF
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -257,7 +257,11 @@ data:
 
         {{- with .tls }}
         tls: {
-          {{- if .custom }}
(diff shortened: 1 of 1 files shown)


caleblloyd:
I think you can just set `.leafnodes.tls.custom="insecure: true"`

caleblloyd:
let me know if that doesn't work and we can revisit

caleblloyd:
It is just a string for the configuration block for that leaf node...  It's essentially an escape hatch for any valid NATS leaf node TLS config

vankosa:
Ok. thanks

PR #544, opened 2022-08-04
Add support for custom TLS to all features

Grendel7:
Custom TLS support was added for leafnodes, but not for any of the other components. But for some more exotic secrets implementations (like secrets-csi-driver), you need the ability to provide your own mounts instead.

This change adds the custom TLS implementation for leafnodes to all other places where TLS is used.


caleblloyd:
Supported in [nats-1.0.0-beta.0](https://github.com/nats-io/k8s/tree/nats-1.0.0-beta.0/helm/charts/nats)

PR #545, opened 2022-08-04
Add support for custom env vars for configuration

Grendel7:
Also use env vars to configure gateway auth, so they can be loaded from secrets instead.

I made a generic snippet for generating the env vars for authentication, and I would _love_ to use it for cluster auth as well, but I can't figure out a way to parameterize them in the cluster peer URLs.

### DIFF
--- a/helm/charts/nats/templates/_helpers.tpl
+++ b/helm/charts/nats/templates/_helpers.tpl
@@ -101,6 +101,26 @@ Return the NATS cluster routes.
 {{- end -}}
 {{- end }}
 
+{{- define "nats.authenticationEnv" }}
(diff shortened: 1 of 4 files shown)


caleblloyd:
Started working on a new helm chart that is much more composable here:

https://github.com/nats-io/k8s/tree/helm-nats-next/helm/charts/nats-next

I think it should be able to do what you are wanting.  It's not released yet but I am hoping to get something out there in the next month or so

caleblloyd:
Supported in [nats-1.0.0-beta.0](https://github.com/nats-io/k8s/tree/nats-1.0.0-beta.0/helm/charts/nats)

PR #548, opened 2022-08-06
[nack helm] add nats.nkey

hhk7734:
https://github.com/nats-io/nack/blob/a36c58fd7ad7d2aeddf17ccd32caa569cdd5826f/cmd/jetstream-controller/main.go#L56

nack has `--nkey` option. but nack helm not support this.
So I added it and checked whether it works, and it works well.


wallyqs (approved):
Thanks!

PR #552, opened 2022-08-11
Update surveyor version

nsurfer:
This updates the Surveyor version in `Chart.yaml`. The file `values.yaml` is left alone because it uses the `latest` tag.
https://github.com/nats-io/k8s/blob/main/helm/charts/surveyor/values.yaml#L10

Also, this deletes some old Surveyor yaml files that we no longer use.


wallyqs (approved):
LGTM

caleblloyd:
Shouldn't we pin the version of surveyor? Especially because we are using `imagePullPolicy: IfNotPresent`

wallyqs:
+1 in the other charts we pin too so easier to understand that image version upgraded a long with the charts

PR #553, opened 2022-08-16
Update surveyor to 0.3.1

nsurfer:


helm/charts/surveyor/Chart.yaml
```diff
@@ -4,4 +4,4 @@ name: surveyor
 description: NATS Monitoring, Simplified.
 type: application
 version: 0.13.2
-appVersion: 0.3.0
+appVersion: 0.3.1
```
wallyqs, line 7: Can you bump the chart version too so that release is triggered?

PR #557, opened 2022-08-25
[nats helm] remove duplicate monitoring port

### LINKED ISSUE #556: The metrics port 7777 is defined two times in the statefulset
First occurence: 
[Container "exporter"](https://github.com/nats-io/k8s/blob/81418cf9f9c0228a767941fac09123dc197389e5/helm/charts/nats/templates/statefulset.yaml#L558)

Second: occurence, maybe wrong, because the *nats* container has no open port 7777
[Container "nats"](https://github.com/nats-io/k8s/blob/81418cf9f9c0228a767941fac09123dc197389e5/helm/charts/nats/templates/statefulset.yaml#L279)

this results in two time series generated from the [prometheus operator](https://github.com/prometheus-operator/prometheus-operator) when using a PodMonitor and and a podMetricsEndpoints of "metrics".

**intended solution:** remove `containerPort: 7777` from the *nats* container spec.

caleblloyd:
Fixes #556

### DIFF
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -276,8 +276,6 @@ spec:
           name: cluster
         - containerPort: 8222
           name: monitor
-        - containerPort: 7777
-          name: {{ .Values.exporter.portName }}
         {{- if .Values.mqtt.enabled }}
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

PR #564, opened 2022-09-02
surveyor: add accounts to monitor metrics

wallyqs:
Example:

```yaml
config:
  servers: "nats://sys:sys@nats:4222"

  expectedServers: 3

  accounts: true
```


caleblloyd (approved):
LGTM

PR #568, opened 2022-09-12
Enable external IPs microk8s

andretadeu:
This PR introduces a way to enable external IPs to access NATS when the K8s cluster does not provide them.

Example:
```bash
helm install my-nats -n my-nats --create-namespace nats/nats --set "nats.requiresExternalIPs.enabled=true" --set "nats.requiresExternalIPs.externalIPs={<External IPs separated by comma>}"
```

### DIFF
--- a/helm/charts/nats/templates/service.yaml
+++ b/helm/charts/nats/templates/service.yaml
@@ -14,7 +14,15 @@ metadata:
 spec:
   selector:
     {{- include "nats.selectorLabels" . | nindent 4 }}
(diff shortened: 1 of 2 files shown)


caleblloyd:
The headless service is required by the stateful set, therefore `clusterIP: none` must not be changed for this service.

The best thing to do would be to add an additional service for your particular needs out-of-band from this chart. This could be done by making your own chart and adding the NATS chart as a dependency.

PR #573, opened 2022-09-21
[kine] add kine-nats development stack

caleblloyd:
While working on [kine](https://github.com/k3s-io/kine/) PRs, I found it helpful to have a development stack that includes NATS

Since it is `k8s` related I thought it would fit well here, but I could make a separate repo for it to if that would be better


wallyqs (approved):
sounds good to me to add here 👍

PR #583, opened 2022-10-12
[nats helm] add opt-in for /healthz liveness and readiness probe

### LINKED ISSUE #576: JetStream Readiness / Liveness probes
Currently we are checking HTTP GET of `/` on the monitoring port for Liveness / Readiness

We _could_ be checking `/healthz?js-server-only=true`:

https://github.com/nats-io/nats-server/blob/main/server/monitor.go#L3049-L3052

If JetStream is enabled, this would check that:

- JetStream has established connection with meta leader
- JetStream is current with meta leader

The main concern here is unintentionally causing a cascading failure.  Example would be a meta leader change followed by it taking a while to catch up.  `/healthz?js-server-only=true` will fail during this, but killing this pod will not "help", it just needs time to catch up

Is this a good idea to enable?  What would reasonable defaults for polling frequency and number of failures be?

/cc @tbeets @wallyqs @derekcollison

caleblloyd:
Resolves #576 

Supported in nats-server versions 2.9.0 and later

Defaults readiness probes to `enabled`, should not affect existing deployments because by default it is the same check as the liveness probe

Adds opt-in feature flag `nats.healthcheck.enableHealthzLivenessReadiness`

This feature flag will be removed and the feature defaulted to true once deemed stable

### DIFF
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -410,12 +410,21 @@ spec:
         #                     #
         #######################
         {{- if .Values.nats.healthcheck }}
+        {{- $serverVersion := (split ":" .Values.nats.image)._1 | default "latest" | regexFind "\\d+(\\.\\d+)?(\\.\\d+)?" | default "2.9.0" }}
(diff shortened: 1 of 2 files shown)


ColinSullivan1 (approved):
LGTM.

PR #586, opened 2022-10-17
Add subjects mapping support in nats chart

OpenGuidou:
According to the doc here: https://docs.nats.io/running-a-nats-service/configuration/configuring_subject_mapping

### DIFF
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -105,7 +105,7 @@ data:
       {{- end }}
     }
     {{- end }}
-    {{- if .Values.nats.subjectsMapping }}
+    {{- if .Values.nats.mappings }}
     ###################################
     #                                 #
     # NATS subjects mapping           #
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -108,7 +108,7 @@ data:
     {{- if .Values.nats.mappings }}
     ###################################
     #                                 #
-    # NATS subjects mapping           #
+    # Mappings                        #
     #                                 #
     ###################################
     mappings: {{ toRawJson .Values.nats.subjectsMapping }}
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -105,6 +105,14 @@ data:
       {{- end }}
     }
     {{- end }}
+    {{- if .Values.nats.subjectsMapping }}
+    ###################################
+    #                                 #
+    # NATS subjects mapping           #
+    #                                 #
+    ###################################
+    mappings: {{ toRawJson .Values.nats.subjectsMapping }}
+    {{- end }}
     {{- if .Values.mqtt.enabled }}
     ###################################
     #                                 #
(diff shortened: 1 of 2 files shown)


helm/charts/nats/values.yaml
```diff
@@ -196,6 +196,35 @@ nats:
...
 
+  # Subject mapping
+  #
+  #   You can find more info at: https://docs.nats.io/running-a-nats-service/configuration/configuring_subject_mapping
+  #
+  # subjectsMapping:
+  #   - source: foo
+  #     destinations:
+  #       - subject: "bar"
+  #   - source: foo.cluster.scoped,
```
skreborn, line 207: Are the commas at the ends of lines supposed to be there?

OpenGuidou: > Are the commas at the ends of lines supposed to be there?

You're right it's an issue in the doc only

OpenGuidou:
@skreborn Is it ok now ?

skreborn:
> @skreborn Is it ok now ?

Oh, me, I'm all good, nothing to complain about, but I unfortunately have no power over moving this along.

wallyqs:
toPrettyJson would escape html so if there are wildcards it would become html entities, maybe toRawJson could work

caleblloyd:
ah good point thanks @wallyqs - yea that would look quite a bit uglier when read out but at least we would not have to maintain the schema in `values.yaml`.  updated my code snippet to `toRawJson`

OpenGuidou:
It seems to linearize the config and the server doesn't like that:

```
###################################
    #                                 #
    # NATS subjects mapping           #
    #                                 #
    ###################################
    mappings: " { foo: bar foo.cluster.scoped: [ { destination: bar.cluster.scoped, weight: 70%, cluster: us-west-1 }, { destination: foobar.cluster.scoped, weight: 30%, cluster: us-east-1 } ] } "
```
nats-server: /etc/nats-config/nats.conf:19:1: interface conversion: interface {} is string, not map[string]interface {}

OpenGuidou:
Thanks for the review.
I updated it based on your comments, and tested it successfully on my side.

helm/charts/nats/templates/configmap.yaml
```diff
@@ -105,6 +105,14 @@ data:
       {{- end }}
     }
     {{- end }}
+    {{- if .Values.nats.subjectsMapping }}
```
caleblloyd, line 108:
```suggestion
    {{- if .Values.nats.mappings }}
```

helm/charts/nats/templates/configmap.yaml
```diff
@@ -105,6 +105,14 @@ data:
...
+    ###################################
+    #                                 #
+    # NATS subjects mapping           #
```
caleblloyd, line 111:
```suggestion
    # Mappings                        #
```

helm/charts/nats/templates/configmap.yaml
```diff
@@ -105,6 +105,14 @@ data:
...
+    #                                 #
+    ###################################
+    mappings: {{ toRawJson .Values.nats.subjectsMapping }}
```
caleblloyd, line 114:
```suggestion
    mappings: {{ toRawJson .Values.nats.mappings }}
```

caleblloyd (commented):
looks good, going to rename `.Values.nats.subjectMappings` to `.Values.nats.mappings` to match the NATS config key exactly

caleblloyd:
Need to run a couple tests then I will add approval if all working

caleblloyd (approved):
LGTM.  Thanks for the contribution!

PR #588, opened 2022-10-19
[nats helm] allow overriding probe paths

caleblloyd:
Allow overriding probe paths, such as:

```
nats.healthcheck.startup.httpGet.path=/
nats.healthcheck.readiness.httpGet.path=/
nats.healthcheck.liveness.httpGet.path=/
```

Also switches probes to use `toYaml` meaning that any supported option for probes that may be added to the future in k8s will be supported


ColinSullivan1 (approved):
LGTM.

wallyqs (approved):

PR #589, opened 2022-10-19
Remove replicas from nack

Joibel:
Replicas doesn't do anything. Nack is not safe run with multiple replicas (as far as I can tell), so remove from values


caleblloyd:
True, it is hard-coded to 1

https://github.com/nats-io/k8s/blob/main/helm/charts/nack/templates/deployment-jetstream-controller.yml#L16

caleblloyd (approved):

PR #590, opened 2022-10-20
Update nats to latest version

Joibel:


caleblloyd (approved):
Thanks.  I removed the chart bump, we will do that as part of the release process

PR #595, opened 2022-11-15
adding hostNetwork and dnsPolicy to helm chart

real-danm:
This PR allows you to configure hostNetwork'ing mode along with a dnsPolicy on the nats statefulset. This can be useful when nats traffic is required to run on specific interfaces attached to a VM.

### DIFF
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -81,6 +81,12 @@ spec:
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName | quote }}
       {{- end }}
+      {{- with .Values.dnsPolicy }}
+      dnsPolicy: {{ toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.hostNetwork }}
+      hostNetwork: {{ toYaml . | nindent 8 }}
+      {{- end }}
       # Common volumes for the containers.
       volumes:
       - name: config-volume
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -81,11 +81,11 @@ spec:
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName | quote }}
       {{- end }}
-      {{- with .Values.dnsPolicy }}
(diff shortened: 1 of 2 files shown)


helm/charts/nats/values.yaml
```diff
@@ -329,6 +329,13 @@ affinity: {}
 ## ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass
 priorityClassName: null
 
+## hostNetwork
+hostNetwork: false
```
caleblloyd, line 333: should be under `nats` key as this only applies to the stateful set

helm/charts/nats/values.yaml
```diff
@@ -329,6 +329,13 @@ affinity: {}
...
+
+## Pod Dns Policy. Default is ClusterFirst
+## ref: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
+dnsPolicy: ClusterFirst
```
caleblloyd, line 337: should be under `nats` key as this only applies to the stateful set

helm/charts/nats/templates/statefulset.yaml
```diff
@@ -81,6 +81,12 @@ spec:
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName | quote }}
       {{- end }}
+      {{- with .Values.dnsPolicy }}
+      dnsPolicy: {{ toYaml . | nindent 8 }}
```
caleblloyd, line 85: this is always a string, so just put the string here, no need for `toYaml`

helm/charts/nats/templates/statefulset.yaml
```diff
@@ -81,6 +81,12 @@ spec:
...
+      {{- end }}
+      {{- with .Values.hostNetwork }}
+      hostNetwork: {{ toYaml . | nindent 8 }}
```
caleblloyd, line 88: this is always a bool, so just put the bool here, no need for `toYaml`

caleblloyd (changes requested):
Thanks for the contribution.  Please add a description and briefly describe what this PR does.

caleblloyd (approved):

caleblloyd:
Looks good to me, @wallyqs could you also take a look?

wallyqs (approved):
LGTM

PR #605, opened 2022-11-24
feat: Allows to activate raft compression

danielcibrao-form3:
According to the change in tha nats-server
https://github.com/nats-io/nats-server/pull/3419, there is the possibility of enabling raft compression in the nats-server. The following PRs gives the possibility of changing that property in the helm release so we can activate it in the nats-server.


danielcibrao-form3:
raft compression is on by default. Closing the PR

PR #613, opened 2022-12-09
[nats helm] fix healthz liveness/readiness probes

caleblloyd:
Fixes an invalid assumption about the health check in the liveness probe to align with https://github.com/nats-io/nats-server/pull/3704

### Server Version `<2.9.0` or  setting of `nats.healthcheck.enableHealthzLivenessReadiness=false`

No Change

### Server Version `>=2.9.0` but `<2.9.10` with setting of `nats.healthcheck.enableHealthzLivenessReadiness=true`

- Liveness Probe:
  - JS Enabled: `/healthz?js-server-only=true&js-enabled=true`
  - JS Disabled: `/healthz?js-server-only=true`

- Readiness Probe:
  - JS Enabled: `/healthz?js-server-only=true&js-enabled=true`
  - JS Disabled: `/healthz?js-server-only=true`

### Server Version `>=2.9.10` with setting of `nats.healthcheck.enableHealthzLivenessReadiness=true`

- Liveness Probe: `/healthz?js-enabled-only=true`
- Readiness Probe: `/healthz?js-server-only=true`

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -8,7 +8,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 0.19.2
+version: 0.19.3
 home: http://github.com/nats-io/k8s
 maintainers:
 - name: Waldemar Quevedo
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

PR #618, opened 2022-12-19
Fixes to support enabling default $SYS account for JS

### LINKED ISSUE #617: Adding a SYS user hides any existing streams
## Defect

Make sure that these boxes are checked before submitting your issue -- thank you!

 - [ ] Included `nats-server -DV` output
 - [x] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve)

#### Versions of `nats-server` and affected client libraries used:
2.9.9

#### OS/Container environment:
Docker, K8s

#### Steps or code to reproduce the issue:

- Create the following config file:

conf.yaml
```
nats:
  image: nats:alpine

  jetstream:
    enabled: true

    fileStorage:
      enabled: true
      size: "1Gi"
      storageDirectory: /data/

cluster:
  enabled: true
  replicas: 3
```

- Install nats using latest helm charts using the command: 
  - `helm install nats nats/nats --values conf.yaml`

- Create a stream using the nats cli

- Create the following config file with basic auth:

config_new.yaml
```
nats:
  image: nats:alpine

  jetstream:
    enabled: true

    fileStorage:
      enabled: true
      size: "1Gi"
      storageDirectory: /data/

cluster:
  enabled: true
  replicas: 3

auth:
  enabled: true
  systemAccount: SYS
  basic:
    noAuthUser: user
    accounts:
      SYS:
        users:
        - user: sys
          password: sys
      \$G:
        users:
        - user: user
          password: user
        jetstream: true
```

- Upgrade nats using latest helm charts using the command: 
  - `helm upgrade nats nats/nats --values conf_new.yaml`

- Check the stream created before the upgrade. The stream does not show up when running: `nats stream ls`

#### Expected result:

The config reloader reloads the new auth config without restarting the nats servers. The stream is still present and can be used. 

#### Actual result:

The nats servers are restarted. The stream is not visible and  also any consumers that created earlier (if any) are not retrievable.

wallyqs:
Fixes #617 

### DIFF
--- a/helm/charts/nats/templates/configmap.yaml
+++ b/helm/charts/nats/templates/configmap.yaml
@@ -471,7 +471,7 @@ data:
             {{- end }}
 
             {{- if .systemAccount }}
-            system_account: {{ .systemAccount }}
+            system_account: {{ .systemAccount | quote }}
             {{- end }}
           {{- end }}
 
@@ -501,7 +501,7 @@ data:
     {{- end }}
 
     {{- with .Values.auth.systemAccount }}
-    system_account: {{ . }}
+    system_account: {{ . | quote }}
     {{- end }}
 
     {{- with .Values.auth.token }}
@@ -578,7 +578,6 @@ data:
       timeout: {{ $.Values.auth.timeout }}
       {{- end }}
     }
-
     accounts: {{- toRawJson . }}
     {{- end }}
 


helm/charts/nats/templates/configmap.yaml
```diff
@@ -573,11 +573,13 @@ data:
     {{- end }}
 
     {{- with .accounts }}
+    {{- if not $.Values.auth.basic.users }}
```
caleblloyd, line 576: What is the goal of this conditional? It is difficult to understand the cases in which it would be true

wallyqs: in case they have both `authorization {}` and `accounts {}` then the `authorization {}` blocks gets repeated twice, I will add a note

caleblloyd: So should the check be `if not $.Values.auth` then?  It seems like there are many cases where the `authorization` block is put in above this if various `$.Values.auth` settings are set

wallyqs: reduced scope for now to be able to set `$SYS` which was not possible due to dollar sign

caleblloyd (approved):
LGTM

PR #620, opened 2022-12-19
Disable config checksum restart by default

wallyqs:
Follow up from #617 

A server can restart with an incompatible config that was not possible to reload and break JS quorum accidentally for example, also reloading of leafnode remotes is supported since v2.9.X so we could opt-out for default now.

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -114,7 +114,7 @@ nats:
 
   # Adds a hash of the ConfigMap as a pod annotation
   # This will cause the StatefulSet to roll when the ConfigMap is updated
-  configChecksumAnnotation: true
+  configChecksumAnnotation: false
 
   # securityContext for the nats container
   securityContext: {}


caleblloyd:
Why would it break JS quorum? If the config is invalid and you are HA with 3+ servers, the 1st one should crash loop because the config is invalid, and the other 2 should stay up with the old config.

Plus this is just kicking the can down the road. Next time k8s upgrades and the underlying nodes are replaced, the pods will restart anyways, and you will then get the crash loop for an invalid config.

I think the only way we should disable the checksum is if we can come up with a way to detect that the config change would be fully and properly handled by the config reloader

wallyqs:
Maybe it is a special case for the sys account, quorum cannot be formed after the restart because the new pod starts with a different sys account so the config is valid so it is unable to boot because of startup checks failing since it cannot find its peers (the `$JSC.>` traffic). The reloader failing to reload would have avoided that, and in case a force restart of all the pods would have been force updated to use the new sys account and they would have been able to find each other since again using the same sys account. 
In other cases it might have helped though as you mention, to detect early when running with an invalid config and avoid all nodes down accidentally so agree to leave it on still.

PR #637, opened 2023-01-17
Add nats-box volumes docs

wallyqs:
Added a note on using the extra volume support for nats-box


neilalexander (approved):
Thanks for this!

PR #650, opened 2023-01-27
feat: create and set Surveyor ServiceAccount

julian-dolce-form3:
Adding the ability to set a `ServiceAccount` for Surveyor.

The chart already had the functionality for generating the name with `"surveyor.serviceAccountName"`, but didn't actually create or set it.


wallyqs (approved):
LGTM, thanks

PR #666, opened 2023-02-13
NACK: Add additionalArgs value to provide to the controller

dmitryabramov-f3:
Allow to specify a dedicated namespace for `nack` to watch instead of the same namespace it's deployed to.
The use case for us is that we have several `nats servers` and corresponding `nacks` running in the same cluster that should be looking in specific namespaces for CRD resources.

UPD: as suggested by @caleblloyd - introducing generic `additionalArgs` value to pass to the controller process.
In our case it will be:
```
additionalArgs:
  - "--namespace=<dedicated namespace>"
```

### DIFF
--- a/helm/charts/nack/Chart.yaml
+++ b/helm/charts/nack/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 - nats
 - jetstream
 - cncf
-version: 0.20.3
+version: 0.20.4
 maintainers:
 - name: Waldemar Quevedo
   url: https://github.com/wallyqs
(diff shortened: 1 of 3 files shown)


caleblloyd:
I think it might be better to call it `additionalArgs: []` in `values.yaml` and pass all of the additional args specified in that array to the controller process

caleblloyd:
Please allow maintainers to edit your PR branch, there is a change I need to push to this

caleblloyd:
> Please allow maintainers to edit your PR branch, there is a change I need to push to this

Actually it may already be enabled, I was trying to use GH Codespaces because editing forks is a lot of steps locally but GH Codespaces may be blocking my push :man_facepalming:

caleblloyd (approved):
I moved `additionalArgs` to `jetstream.additionalArgs`. LGTM

PR #669, opened 2023-02-16
make emptyDir volumes configurable

monotek:
This pr makes emptyDir volumes pid & advertiseconfig configurable.

Normal "emptyDir" volume stays the default but now you can also set a sizeLimit for the emptyDir for example or even configure to use another volumetype, like a pvc.

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -8,7 +8,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 0.20.0
+version: 0.19.10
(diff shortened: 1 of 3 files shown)


helm/charts/nats/Chart.yaml
```diff
@@ -8,7 +8,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 0.19.10
+version: 0.20.0
```
wallyqs, line 11:
```suggestion
version: 0.19.10
```

wallyqs (approved):

PR #670, opened 2023-02-16
NATS: Alternate [Optional] Jetstream store_dir and max_file Configuration

david-gracey-form3:
In the case where Jetstream is enabled but `fileStorage` is not enabled (e.g. when using `hostPath` storage), optionally allow specifying `store_dir` and `max_file` via `values.yaml`. These will be directly set in the `jetstream` section of `nats.conf` configuration file.

Also allow for specifying additional environment variables (`.Values.nats.extraEnv`) to the `nats-server` container for convenience.

### DIFF
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -333,6 +333,9 @@ spec:
         - name: GOMEMLIMIT
           value: {{ .Values.nats.gomemlimit | quote }}
         {{- end }}
+        {{- if .Values.nats.extraEnv }}
+        {{- toYaml .Values.nats.extraEnv | nindent 8 }}
+        {{- end }}
 
         {{- if .Values.nats.jetstream.enabled }}
         {{- with .Values.nats.jetstream.encryption }}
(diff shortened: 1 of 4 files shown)


helm/charts/nats/templates/statefulset.yaml
```diff
@@ -333,6 +333,9 @@ spec:
         - name: GOMEMLIMIT
           value: {{ .Values.nats.gomemlimit | quote }}
         {{- end }}
+        {{- if .Values.extraEnvs }}
+        {{- toYaml .Values.extraEnvs | nindent 8 }}
```
wallyqs, line 337: How about moving this one to be under `.Values.nats.extraEnvs` since it only applies to that container?

caleblloyd: I recommend `.Values.nats.extraEnv` with no trailing s, so that we stay as close to possible as the schema and `env:` has no trailing s

david-gracey-form3: LGTM

wallyqs (approved):
LGTM

PR #671, opened 2023-02-17
Allow possibility to pass desired flags for prometheus exporter

faisal-khalique:
It should be possible to select the metrics to fetch from the prometheus exporters. This PR allows for `flags` as describe [here](https://github.com/nats-io/prometheus-nats-exporter#usage) to be passed through `values` and not take the default.

Sample snippet for prometheus exporter configuration when requiring desired flags
```yaml
# Prometheus NATS Exporter configuration.
exporter:
  enabled: true
  image:
    repository: natsio/prometheus-nats-exporter
    tag: 0.10.1
    pullPolicy: IfNotPresent
    # registry: docker.io

  portName: metrics
  securityContext: {}
  resources: {}
  # an array of flags to pass to exporter. See https://github.com/nats-io/prometheus-nats-exporter#usage
  flags:
  - -connz
  - -routez
  # Prometheus operator ServiceMonitor support. Exporter has to be enabled
  serviceMonitor:
    enabled: false
    ## Specify the namespace where Prometheus Operator is running
    ##
    # namespace: monitoring
    labels: {}
    annotations: {}
    path: /metrics
    # interval:
    # scrapeTimeout:
```

### DIFF
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -593,8 +593,8 @@ spec:
         {{- if .Values.gateway.enabled }}
         - -gatewayz
         {{- end }}
-        {{- end }}
         - http://localhost:8222/
+        {{- end }}
         ports:
         - containerPort: 7777
           name: {{ .Values.exporter.portName }}
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -575,10 +575,8 @@ spec:
         resources:
           {{- toYaml .Values.exporter.resources | nindent 10 }}
         args:
-        {{- if .Values.exporter.flags }}
-        {{- range .Values.exporter.flags }}
-        - {{ . }}
-        {{- end }}
+        {{- if .Values.exporter.args }}
+        {{- toYaml .Values.exporter.args | nindent 8 }}
         {{- else }}
         - -connz
         - -routez
--- a/helm/charts/nats/templates/statefulset.yaml
+++ b/helm/charts/nats/templates/statefulset.yaml
@@ -575,6 +575,11 @@ spec:
         resources:
           {{- toYaml .Values.exporter.resources | nindent 10 }}
         args:
+        {{- if .Values.exporter.flags }}
+        {{- range .Values.exporter.flags }}
+        - {{ . }}
+        {{- end }}
+        {{- else }}
         - -connz
         - -routez
         - -subz
@@ -589,6 +594,7 @@ spec:
         {{- end }}
         {{- if .Values.gateway.enabled }}
         - -gatewayz
+        {{- end }}
         {{- end }}
         - http://localhost:8222/
         ports:
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -610,8 +610,9 @@ exporter:
   portName: metrics
   securityContext: {}
   resources: {}
-  # an array of flags to pass to exporter. See https://github.com/nats-io/prometheus-nats-exporter#usage
-  flags: ~
+  # override the default args passed to the exporter
+  # see https://github.com/nats-io/prometheus-nats-exporter#usage
+  args: []
   # Prometheus operator ServiceMonitor support. Exporter has to be enabled
   serviceMonitor:
     enabled: false
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -610,6 +610,8 @@ exporter:
   portName: metrics
   securityContext: {}
   resources: {}
+  # an array of flags to pass to exporter. See https://github.com/nats-io/prometheus-nats-exporter#usage
+  flags: ~
   # Prometheus operator ServiceMonitor support. Exporter has to be enabled
   serviceMonitor:
     enabled: false
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -612,6 +612,7 @@ exporter:
   resources: {}
   # override the default args passed to the exporter
   # see https://github.com/nats-io/prometheus-nats-exporter#usage
+  # make sure to pass HTTP monitoring port URL as last arg, e.g ["-connz", "http://localhost:8222/"]
   args: []
   # Prometheus operator ServiceMonitor support. Exporter has to be enabled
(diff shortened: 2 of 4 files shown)


faisal-khalique:
This PR should allow for the new flags to be passed (https://github.com/nats-io/prometheus-nats-exporter/pull/208)

Notwithstanding the new flags, it should be possible to pass own flags.

/ @wallyqs

helm/charts/nats/templates/statefulset.yaml
```diff
@@ -575,6 +575,11 @@ spec:
         resources:
           {{- toYaml .Values.exporter.resources | nindent 10 }}
         args:
+        {{- if .Values.exporter.flags }}
+        {{- range .Values.exporter.flags }}
+        - {{ . }}
+        {{- end }}
+        {{- else }}
```
caleblloyd, line 582:
```suggestion
        {{- if .Values.exporter.args }}
        {{- toYaml .Values.exporter.args | nindent 8 }}
        {{- else }}
```

helm/charts/nats/values.yaml
```diff
@@ -610,6 +610,8 @@ exporter:
   portName: metrics
   securityContext: {}
   resources: {}
+  # an array of flags to pass to exporter. See https://github.com/nats-io/prometheus-nats-exporter#usage
+  flags: ~
```
caleblloyd, line 614:
```suggestion
  # override the default args passed to the exporter
  # see https://github.com/nats-io/prometheus-nats-exporter#usage
  args: []
```

helm/charts/stan/templates/statefulset.yaml
```diff
@@ -307,12 +307,18 @@ spec:
             {{- toYaml .Values.exporter.securityContext | nindent 12 }}
           {{- end }}
           args:
+          {{- if .Values.exporter.flags }}
+          {{- range .Values.exporter.flags }}
+          - {{ . }}
+          {{- end }}
+          {{- else }}
```
caleblloyd, line 314: {{- if .Values.exporter.args }}
        {{- toYaml .Values.exporter.args | nindent 8 }}
        {{- else }}

helm/charts/stan/values.yaml
```diff
@@ -273,6 +273,8 @@ exporter:
   pullPolicy: IfNotPresent
   securityContext: {}
   resources: {}
+  # an array of flags to pass to exporter. See https://github.com/nats-io/prometheus-nats-exporter#usage
+  flags: ~
```
caleblloyd, line 277:
```suggestion
  # override the default args passed to the exporter
  # see https://github.com/nats-io/prometheus-nats-exporter#usage
  args: []
```

caleblloyd (changes requested):

caleblloyd:
I changed `flags` -> `args` because we try to match the k8s schema exactly wherever we can

faisal-khalique:
> I changed `flags` -> `args` because we try to match the k8s schema exactly wherever we can

I was in doubt as well. I used `flags` to differentiate from the fact the the `URL` (part of arguments) was not getting changed with this.

caleblloyd:
oh I missed that detail...  now I see that `flags` may make more sense :sweat_smile: 

is there ever a case in which the URL would need to get overridden?  are you trying to add flags or take some away?

if you are only adding flags we could call it `additionalArgs` and tack them onto the end

caleblloyd:
I don't like the idea of making it `flags` now, then someone else has the requirement to change the URL and we have to make it `args` later.

I'd rather just remove the URL and if you want to wholesale replace the `args` then the URL would be part of that replacement...

helm/charts/stan/templates/statefulset.yaml
```diff
@@ -307,12 +307,16 @@ spec:
...
+          {{- toYaml .Values.exporter.args | nindent 10 }}
+          {{- else }}
           - -connz
           - -routez
           - -subz
           - -varz
           - -channelz
           - -serverz
+          {{- end }}
           - http://localhost:8222/
```
caleblloyd, line 320:
```suggestion
```

faisal-khalique:
I don't see. Since, exporter is running is side car, it will be localhost only. And well for the port numbers, I doubt anyone would go that way.

helm/charts/nats/templates/statefulset.yaml
```diff
@@ -590,6 +593,7 @@ spec:
         {{- if .Values.gateway.enabled }}
         - -gatewayz
         {{- end }}
+        {{- end }}
         - http://localhost:8222/
```
caleblloyd, line 597:
```suggestion
```

caleblloyd (commented):
wholesale arg replacement

caleblloyd:
Never say never with helm, we get requests for just about every single option :smile: 

Makes me wish it supported Kustomize out of the box most days

helm/charts/nats/values.yaml
```diff
@@ -610,6 +610,9 @@ exporter:
   portName: metrics
   securityContext: {}
   resources: {}
+  # override the default args passed to the exporter
+  # see https://github.com/nats-io/prometheus-nats-exporter#usage
```
caleblloyd, line 614:
```suggestion
  # override the default args passed to the exporter
  # see https://github.com/nats-io/prometheus-nats-exporter#usage
  # make sure to pass HTTP monitoring port URL as last arg, e.g ["-connz", "http://localhost:8222/"]
```

helm/charts/stan/values.yaml
```diff
@@ -273,6 +273,9 @@ exporter:
   pullPolicy: IfNotPresent
   securityContext: {}
   resources: {}
+  # override the default args passed to the exporter
+  # see https://github.com/nats-io/prometheus-nats-exporter#usage
```
caleblloyd, line 277:
```suggestion
  # override the default args passed to the exporter
  # see https://github.com/nats-io/prometheus-nats-exporter#usage
  # make sure to pass HTTP monitoring port URL as last arg, e.g ["-connz", "http://localhost:8222/"]
```

caleblloyd (commented):
hint in values file

caleblloyd:
Suggestion for wholesale arg replacement with a hint to pass the URL work?

caleblloyd (approved):
LGTM, @wallyqs what do you think?

wallyqs (approved):
LGTM

PR #673, opened 2023-02-21
Add configurable test image in chart

jsecchiero:
make the nats box image used in tests from static to configurable in the helm chart

### DIFF
--- a/helm/charts/nats/templates/tests/test-request-reply.yaml
+++ b/helm/charts/nats/templates/tests/test-request-reply.yaml
@@ -10,7 +10,7 @@ metadata:
 spec:
   containers:
   - name: nats-box
-    image: {{ .Values.natsbox.image }}
+    image: {{ include "nats.image" .Values.natsbox.image }}
     env:
     - name: NATS_HOST
       value: {{ template "nats.fullname" . }}


helm/charts/nats/templates/tests/test-request-reply.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
 spec:
   containers:
   - name: nats-box
-    image: synadia/nats-box
+    image: {{ .Values.natsbox.image }}
```
joschiwald, line 13: this will not render correctly since chart version 0.19.0, there was a huge change with images (see https://github.com/nats-io/k8s/commit/5c319a5aa0ad54c50af17614535ca4783308308a#diff-bd1bd24003c063bee87a6774fc4652b2211cc5ebb40164168ef644e8ae02decd)

jsecchiero: you are right i made a modification based on a different version. Updated!

jsecchiero: @joschiwald Is there anything else that you think I should do?

caleblloyd (approved):

helm/charts/nats/templates/tests/test-request-reply.yaml
joschiwald: Now it looks good to me. 👍 

Just the "nats.fixImage" is missing, but since this is just a test file, I think we could ignore that backward compatibility thing here.

PR #683, opened 2023-03-03
Bump startup timeout to be longer by default

wallyqs:
When having a larger number of streams it was seen  this 5m timeout causing extra restarts which caused systems to take longer than needed to catchup.


caleblloyd (approved):
this pushes startup probe total time to 15m (worst case scenario) - LGTM

PR #684, opened 2023-03-08
Bump initial delay seconds to 90

wallyqs:
Wait longer to ensure that the JetStream server has at least attempted to start and avoid getting an early ok healthz.

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -95,7 +95,7 @@ nats:
     startup:
       enabled: true
 
-      initialDelaySeconds: 10
+      initialDelaySeconds: 90
       timeoutSeconds: 5
       periodSeconds: 10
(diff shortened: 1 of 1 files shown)


caleblloyd:
https://github.com/nats-io/nats-server/blob/a181e90c3926c14503ddb139145751cbe9570cf1/server/monitor.go#L3064-L3076

There should not be a case in which an early Healthz can come through here

wallyqs:
wonder if the early healthz ok exit is coming from here: https://github.com/nats-io/nats-server/blob/a181e90c3926c14503ddb139145751cbe9570cf1/server/monitor.go#L3086-L3112
recovered ok but js.cluster not yet ready so healthz is ok until cluster has state.

caleblloyd:
Can we close this one?

Issue #398, opened 2022-01-10
Provide CRDs in `nack` helm chart


manuelottlik:
Hey guys,

when installing the `nack` helm chart I figured that the [CRDs for the JetStream controller](https://github.com/nats-io/nack/blob/main/deploy/crds.yml) are not installed. This is possible in an helm chart by providing a `crds` folder next to `templates` ([see helm docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#install-a-crd-declaration-before-using-the-resource)).

I think the `nack` helm chart should provide these CRDs. Do you agree? I would be happy to provide a PR for this.

wallyqs:
Thanks @manuelottlik, they used to be part of the Helm chart but we took them out because that meant that updates to the CRDs were not being applied and these CRDs can have new fields fairly often.  Managing CRDs via Helm has many gotchas as described in the following discussion from Helm maintainers, so for now decided to keep them out of the repo: https://github.com/helm/community/blob/main/hips/hip-0011.md#helm-is-still-working-on-thus-but-we-cannot-solve-it-alone

manuelottlik:
Thank you for the quick reply, @wallyqs. There are two things I would like to say:

1. Couldn't we just provide the CRDs in the `crds` folder and give a migration guide whenever there is a new version with changed definitions? The migration guide could contain a `kubectl apply -f ...` link that users of the helm chart could copy to update their CRDs in their cluster. Since users of the helm chart have to do this anyways in order to update this helm chart, it feels like the best UX for me.
2. If CRDs will not be included, this should be stated somewhere in the documentation / README of this helm chart. A link to the CRD file etc. would be nice. New versions that require updated CRDs should also have some kind of migration guide. I would open a PR if you want me to.

What do you think?

wallyqs:
I think both sounds good, just noticed that they are not mentioned in the readme, agree that we need a migration guide / better strategy for handling updates to the CRDs as well...  For a bit there used to be a copy of the CRDs both in these repo and in the nack repo but eventually only left this copy since there was some confusion with duplication: https://github.com/nats-io/nack/blob/main/deploy/crds.yml

manuelottlik:
Hey @wallyqs, of course:
* https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack
* https://github.com/traefik/traefik-helm-chart/tree/master/traefik
* https://github.com/grafana/helm-charts/tree/main/charts/agent-operator
[...]

PR #694, opened 2023-04-03
Copy CRDs into the `crds` directory of the NACK chart

dfr-exnaton:
## Summary

Based on the discussions in https://github.com/nats-io/k8s/issues/398, this PR copies the [required CRDs](https://github.com/nats-io/nack/blob/main/deploy/crds.yml) for NACK into the special `crds` directory of NACK helm chart and updates the README as well as the tests. Drawback is the increased maintenance in keeping the CRDs in sync with the `nack` repo. 

## Motivation

In my opinion, providing a caveat that CRDs have to be manually upgraded as part of a chart upgrade is necessary also when no `crds` directory is used. Thus its more user friendly if CRDs are provided as part of the chart. Not having an extra step to manually install & maintain CRDs when using this chart is also super helpful for GitOps users. Flux for example specifically has [a special configuration object](https://github.com/fluxcd/helm-controller/blob/main/docs/api/helmrelease.md#helm.toolkit.fluxcd.io/v2beta1.CRDsPolicy) that allows to reinstall CRDs as part of an chart upgrade.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,14 +6,14 @@ on:
     branches:
       - main
     paths:
-      - "helm/charts/**"
-      - "!helm/charts/index.yaml"
-      - ".github/workflows/test.yaml"
+      - 'helm/charts/**'
+      - '!helm/charts/index.yaml'
+      - '.github/workflows/test.yaml'
   pull_request:
     paths:
-      - "helm/charts/**"
-      - "!helm/charts/index.yaml"
-      - ".github/workflows/test.yaml"
+      - 'helm/charts/**'
+      - '!helm/charts/index.yaml'
+      - '.github/workflows/test.yaml'
 
 jobs:
   lint-test:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,14 +6,14 @@ on:
     branches:
       - main
     paths:
-      - 'helm/charts/**'
-      - '!helm/charts/index.yaml'
-      - '.github/workflows/test.yaml'
+      - "helm/charts/**"
+      - "!helm/charts/index.yaml"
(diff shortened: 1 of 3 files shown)


ChristianGmw:
Could this PR be merged please. Very useful when you deploy everything via GitOps

1995parham:
What do you think about moving the chart into the main project and then push releases here? Then we have everything on the sample place and make updating things easier.

svenfoo:
Can this be merged, or alternatively solved in some another way? Having to maintain a copy of the CRDs is tedious and error-prone.

dfr-exnaton:
Would love to update the PR (update CRD if necessary and fix merge conflicts) after a maintainer gives a thumbs up on this rather small, non-breaking change.

vvinnich:
@dfr-exnaton what do you think to add
```
annotations:
  helm.sh/resource-policy: keep
  ```

dfr-exnaton:
> @dfr-exnaton what do you think to add
> 
> ```
> annotations:
>   helm.sh/resource-policy: keep
> ```

This PR just copied over the crds to include them in the chart, nothing was modified. Do you have an example where this annotation would help? Because I could also see issues when keeping resources around...

dfr-exnaton:
@caleblloyd would be great if this minor PR can either get a thumbs up or thumbs down. If positive, I would obviously update the README and the crds to be in sync (after more than one year...)

ChristianGmw:
I cannot understand why no maintainer answeres here? Is NATS so badly maintained?

vavsab:
A bit frustrating that in 2024 we still need to copy-paste CRDs while many other charts include them by default.

samuelattwood:
CRDs are now included in the `0.28.0` NACK chart.

PR #696, opened 2023-04-03
fix(surveyor): configmap name as string

Kriys94:
For some reason I got the following error:

```
error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.spec.volumes[0].configMap): invalid type for io.k8s.api.core.v1.ConfigMapVolumeSource: got "string", expected "map"
```

Putting the ConfigMap name under quote solves the issue


Kriys94:
Your suggestion works too. Thanks and I let you merge

caleblloyd (approved):

PR #700, opened 2023-04-11
host helm charts from gh-pages branch

caleblloyd:
Host Helm Charts from the `gh-pages` branch so that the `main` branch can be protected


wallyqs:
I think LGTM, let's give it a try.

wallyqs (approved):

PR #706, opened 2023-04-17
Update NATS helm chart to v2.9.16

bruth:


wallyqs (approved):

caleblloyd:
@bruth Dependabot will do this automatically, and it will auto-bump the chart version too.  I think it checks daily

bruth:
@caleblloyd Good to know thanks, I was over-eager to get it updated.

PR #717, opened 2023-05-09
[nats helm 1.x] fix JS mount

### LINKED ISSUE #716: max_file_store unset in nats.conf
I've just deployed `1.0.0-beta.1` but it doesn't seem to set `max_file_store` in the nats.conf using the following values:
```
config:
  cluster:
    enabled: true
    replicas: 3
  jetstream:
    enabled: true
    fileStore:
      enabled: true
        pvc:
          enabled: true
          size: 25Gi
```

I can set the value manually via `config.jetstream.fileStore.maxSize` but the comment in values.yaml states it should be set as the total available PVC size by default?

caleblloyd:
Fixes #716 

In `1.0.0-beta.2` and below  we are mounting to `/data/jetstream` and the store dir is `/data`

So it looks like NATS Server is detecting the size of `/data` which is outside of the PVC mount

This is going to be a problem when upgrading from 0.x anyways, because in 0.x we are mounting to `/data` and the store dir is `/data`

This updates the mount dir and store dir to match and default to `/data`

There is a one-time-fix initContainer that will be released as part of `1.0.0-beta.3` and removed after that

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 1.0.0-beta.2
+version: 1.0.0-beta.3
 home: http://github.com/nats-io/k8s
 maintainers:
 - email: info@nats.io
--- a/helm/charts/nats/README.md
+++ b/helm/charts/nats/README.md
@@ -15,6 +15,9 @@ helm repo add nats https://nats-io.github.io/k8s/helm/charts/
 helm upgrade --install nats nats/nats --devel
 ```
 
+## 1.x Beta Upgrade Notes
+- If using `1.0.0-beta.2` or lower with JetStream enabled, upgrade to `1.0.0-beta.3` first
+
 ## Values
 
 There are a handful of explicitly defined options which are documented with comments in the [values.yaml](values.yaml) file.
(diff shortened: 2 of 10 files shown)


samuelattwood (approved):
LGTM

caleblloyd:
Going further down the rabbit hole - it turns out that the auto-disk calculation is 75% of available space, but it's got an issue because it will shrink after restarting

https://github.com/nats-io/nats-server/issues/4140

So it seems that the best path forward right now is to adopt the 0.x behavior of defaulting `max_file_storage` to the PVC size.  Although there is still the option to override it with `maxSize`.

It is not all that useful to be set to the PVC Size, because the actual disk is smaller when formatted.  For example a 10Gi disk has 9.7Gi usable after formatting.  So a `max_file_storage` of 10Gi will never be hit, the disk will fill up first.  But it is the same behavior that 0.x has today so I digress

caleblloyd:
Alright at least the logs look right now :smile: 

```
[15] 2023/05/09 20:01:53.040642 [INF] ---------------- JETSTREAM ----------------
[15] 2023/05/09 20:01:53.040645 [INF]   Max Memory:      0 B
[15] 2023/05/09 20:01:53.040648 [INF]   Max Storage:     10.00 GB
[15] 2023/05/09 20:01:53.040650 [INF]   Store Directory: "/data/jetstream"
[15] 2023/05/09 20:01:53.040651 [INF] -------------------------------------------
```

PR #735, opened 2023-06-16
ConfigMap annotations

oldgiova:
I'm using helm-hook annotations to install `nats` as a Sub-Chart of another application Chart:
```
helm.sh/hook: pre-install
```

The issue with this approach is that the statefulset manifest requires the `nats-config` ConfigMap to mount; but this is not available at the `pre-install` stage. 

With this PR I would like to add annotations to the `ConfigMap` as well as the other resources.

Thanks!


oldgiova:
Thanks @caleblloyd the 1.x looks really promising, thanks!

PR #743, opened 2023-06-23
[helm nats 1.x] 0.x -> 1.x upgrade guide

### LINKED ISSUE #712: NATS Helm 0.x -> 1.x Upgrade Path
Develop an upgrade path from the 0.x Helm Chart to 1.x.  The upgrade has 2 goals:

1. Ensure that an existing 0.x StatefulSet can start
    StatefulSets have a number of immutable fields such as the Headless Service, Selector Labels, and PVC Templates.
    Provide instructions or scripting to help users configure the StatefulSet so that the immutable fields can remain un-changed.
2. Migrate values to new format
    The `values.yaml` format has changed significantly.
    Provide instructions or scripting to help show users how to translate values from 0.x to 1.x
    Document incompatible features and default values that are different between 0.x and 1.x

caleblloyd:
Resolves #712

Add an upgrade guide for transitioning from the 0.x -> 1.x Helm Chart

### DIFF
--- a/helm/charts/nats/UPGRADING.md
+++ b/helm/charts/nats/UPGRADING.md
@@ -1,0 +1,136 @@
+# Upgrading from 0.x to 1.x
+
+Instructions for upgrading an existing `nats` 0.x release to 1.x.
+
+## Rename Immutable Fields
+
+There are a number of immutable fields in the NATS Stateful Set and NATS Box deployment.  All 1.x `values.yaml` files targeting an existing 0.x release will require some or all of these settings:
+
+```yaml
+config:
+  # required if using JetStream file storage
+  jetstream:
+    # make sure to uncomment the next line if using JetStream file storage
+    # enabled: true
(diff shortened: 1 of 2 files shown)


helm/charts/nats/UPGRADING.md
```diff
@@ -0,0 +1,136 @@
...
+    # set other PVC options here to make it match 0.x, refer to values.yaml for schema
+
+# required
+statefulSet:
+  patch:
+  - op: remove
+    path: /spec/selector/matchLabels/app.kubernetes.io~1component
+
+# required
+headlessService:
```
wallyqs, line 37: I think we need a note on the headless and cluster services names for when creating certs (need two pair of certs one for each name).

caleblloyd: Added TLS considerations to `README.md` and a comment under the service in `UPGRADING.md`

PR #746, opened 2023-06-23
[helm nats 1.x] config.cluster.routeURLs options

### LINKED ISSUE #744: [helm nats 1.x] Add Route FQDN and Authorization
Add Route FQDN and Authorization to `config.cluster`

caleblloyd:
Resolves #744 

Adds `config.cluster.routeURLs` section to more easily customize route URLs

The `user` and `password` settings go multiple places - both in the route URLs and the `authorization` block - so it makes sense to add a defined setting for those

The `useFQDN` setting is present in 0.x and may be a part of SANs for Route TLS certificates already

### DIFF
--- a/helm/charts/nats/README.md
+++ b/helm/charts/nats/README.md
@@ -183,43 +183,13 @@ Anything in `values.yaml` can be templated:
     yaml template
   ```
 
-Example - add cluster authorization:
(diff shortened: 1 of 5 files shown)


wallyqs (approved):
LGTM

PR #758, opened 2023-07-06
[helm nats 1.x] remove tls.ca options

### LINKED ISSUE #757: [nats helm 1.x] Remove TLS Secret CA options
TLS Secret CA is not explicitly defined in `kubernetes.io/tls`  Secret Type

```
	// SecretTypeTLS contains information about a TLS client or server secret. It
	// is primarily used with TLS termination of the Ingress resource, but may be
	// used in other types.
	//
	// Required fields:
	// - Secret.Data["tls.key"] - TLS private key.
	//   Secret.Data["tls.crt"] - TLS certificate.
	// TODO: Consider supporting different formats, specifying CA/destinationCA.
```

Sometimes issuing CAs are bundled with certs, such as `ca.crt` in the case of CertManager.  But that does not mean this is a CA that should be used for validating client connections, it is simply the CA for the cert bundle.

Therefore we should not have an easy path to make this CA be the `ca_file` in the TLS section

caleblloyd:
Resolves #757

### DIFF
--- a/helm/charts/nats/files/config/tls.yaml
+++ b/helm/charts/nats/files/config/tls.yaml
@@ -3,9 +3,5 @@
 {{- $dir := trimSuffix "/" .dir }}
 cert_file: {{ printf "%s/%s" $dir .cert }}
 key_file: {{ printf "%s/%s" $dir .key }}
-{{- if .ca }}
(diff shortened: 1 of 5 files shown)


wallyqs (approved):
LGTM

philpennock (approved):

PR #762, opened 2023-07-14
[helm nats 1.x] add appProtocol to services

### LINKED ISSUE #761: NATS Helm Chart 1.x support for k8s Service Application Protocol
I recently installed istio and discovered that my applications had difficulty connecting to nats when injecting istio sidecar.

After some experimenting and chatting with @caleblloyd on Slack, we found out that we need to support [`appProtocol`](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol).

The following works

```
 ports:
  - appProtocol: tcp
    name: nats
    port: 4222
    protocol: TCP
    targetPort: nats
```

caleblloyd:
Resolves #761


wallyqs (approved):
LGTM

PR #763, opened 2023-07-14
[helm nats 1.x] add tlsCA option

caleblloyd:
Adds a root level `tlsCA` option that can mount a CA bundle from a ConfigMap or Secret

If this option is supplied, it will use the CA bundle in all NATS Server `tls` blocks, and all `nats-box` contexts

Meant to be used in conjunction with something like [trust-manager](https://cert-manager.io/docs/projects/trust-manager/)


wallyqs (approved):
LGTM

helm/charts/nats/files/nats-box/deployment/container.yaml
```diff
@@ -26,12 +26,17 @@ command:
 args:
 - trap true INT TERM; sleep infinity & wait
 volumeMounts:
+# contexts secret
 - name: contexts
   mountPath: /etc/nats-contexts
+# contents secret
```
philpennock, line 32: Should this be within the template if below? Similarly for pid stuff below?

philpennock (approved):
Everything substantive looks good

helm/charts/nats/files/nats-box/deployment/container.yaml
caleblloyd: I've been in the habit of putting comments at the beginning of conditional blocks, because it helps me see balancing of the `{{- if` and `{{- end` braces between the comments

PR #806, opened 2023-10-02
[nats helm] set probe scheme based on monitor tls

### LINKED ISSUE #805: NATS Chart Probes not Using HTTPS when Monitor TLS is Enabled
Hey!

K8S version: **1.25.3**
Nats chart version: **nats-v1.1.0**

**Defect:**
When `config.monitor.tls` is enabled, probes are generated with the default `httpGet.scheme` "`HTTP`", but it should be "`HTTPS`".

**Fix suggestion:**
In '_files/stateful-set/nats-container.yaml_', add the following to each probe

```
{{- if .Values.config.monitor.tls.enabled }}
scheme: HTTPS
{{- end }}
```

Thanks.

caleblloyd:
Resolves #805

### DIFF
--- a/helm/charts/nats/files/stateful-set/nats-container.yaml
+++ b/helm/charts/nats/files/stateful-set/nats-container.yaml
@@ -33,11 +33,15 @@ lifecycle:
       - nats-server
       - -sl=ldm=/var/run/nats/nats.pid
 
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

Issue #790, opened 2023-08-21
[nats helm] Global Image Pull Secrets


caleblloyd:
Consider adding a global setting for image pull secrets that will affect both nats and natsBox

PR #814, opened 2023-11-01
[nats helm] Allow specifying pullSecrets

adberger:
Allow the use of `pullSecrets` to pull from a different registry than default (https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/).

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 1.1.3
+version: 1.2.0
 home: http://github.com/nats-io/k8s
 maintainers:
 - email: info@nats.io
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 1.2.0
+version: 1.1.3
 home: http://github.com/nats-io/k8s
 maintainers:
 - email: info@nats.io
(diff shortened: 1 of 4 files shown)


helm/charts/nats/values.yaml
```diff
@@ -6,6 +6,12 @@ global:
     # global image pull policy to use for all container images in the chart
     # can be overridden by individual image pullPolicy
     pullPolicy:
+    ## Optionally specify an array of imagePullSecrets.
+    ## Secrets must be manually created in the namespace.
+    ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+    ##
+    # pullSecrets:
+    #   - myRegistrKeySecretName
```
caleblloyd, line 14: Prefer setting to empty value as opposed to comment for syntax highlighting, and so you can search for the value on artifact hub

```suggestion
    # global image pull secrets to use for all pod templates in the chart
    # https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
    pullSecrets: []
```

helm/charts/nats/Chart.yaml
```diff
@@ -6,7 +6,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 1.1.3
+version: 1.2.0
```
caleblloyd, line 9: WIll put in follow-up PR that adds a test and bumps

```suggestion
version: 1.1.3
```

caleblloyd:
Resolves #790

caleblloyd:
Thanks for the contribution!

PR #815, opened 2023-11-01
tests for global.image.pullSecretNames

caleblloyd:
Tests for #814 

Changes un-released `global.image.pullSecrets` to `global.image.pullSecretNames`

This is because `imagePullSecrets` takes an array of `{name: secretName}` objects

I thought it was better to be more specefic that we are only taking the `secretName` portion in `pullSecretNames`, and we are adding the `name: ` key

This is consistent with the use of `secretName` many other places in the chart values

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -36,7 +36,7 @@ jobs:
           python-version: 3.9
 
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.4.0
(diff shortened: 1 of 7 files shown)


adberger:
> /cc @adberger

I'm fine with that, thanks.
Just FYI the other helm charts also don't seem to template `name` automatically.

wallyqs (approved):

samuelattwood (approved):

PR #826, opened 2023-11-17
Prometheus Exporter Port Name

rustik666:
## What this PR does / why we need it
In this PR we add ability to rewrite Prometheus exporter container port name to be able to avoid hardcoded `prom-metrics` value.

## Which issue this PR fixes
For our particular case, we use internal rule to name all ports targets of `ClusterPodMonitor` as `metrics` with `NetworkPolicy` where we allow `gmp-system` to collect metrics from named port.

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 1.1.5
+version: 1.1.6
 home: http://github.com/nats-io/k8s
 maintainers:
 - email: info@nats.io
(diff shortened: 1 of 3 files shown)


x-cray (approved):

caleblloyd:
Already possible with:

```yaml
promExporter:
  enabled: true
  patch:
  - op: replace
    path: /ports/0/name
    value: your-port-name

  # if you are also using the pod monitor, do this too:
  podMonitor:
    enabled: true
    patch:
    - op: replace
      path: /spec/podMetricsEndpoints/0/port
      value: your-port-name
```

rustik666:
> Already possible with:
> 
> ```yaml
> promExporter:
>   enabled: true
>   patch:
>   - op: replace
>     path: /ports/0/name
>     value: your-port-name
> 
>   # if you are also using the pod monitor, do this too:
>   podMonitor:
>     enabled: true
>     patch:
>     - op: replace
>       path: /spec/podMetricsEndpoints/0/port
>       value: your-port-name
> ```

@caleblloyd thank you very much!

PR #832, opened 2023-11-23
[surveyor helm] add support for nkey in the same way as credentials

0xflatheadfred:
Add support for specification of an nkey file for auth. Everything is exactly identical to credentials.


caleblloyd (approved):
LGTM

PR #839, opened 2023-12-11
Bump nats from 2.10.5-alpine to 2.10.7-alpine in /helm/charts/nats

svenfoo:


svenfoo:
Usually this seems to be done by `dependabot`. But I haven't seen such a PR and we would like to start using `nats-server 2.10.7`.

caleblloyd (approved):
Thanks for the heads up.  Sure enough, Dependabot jobs started failing Dec 1st :frowning: #840 opened to track

PR #849, opened 2023-12-29
Update reloader image to latest patch version

sboulkour:
This patch fixes vulnerability scan finding critical vulns. This prevent the image to be imported into hardened kubernetes env and associated local registry.

Issue #840, opened 2023-12-11
Dependabot Failing


caleblloyd:
### What version were you using?

Dependabot jobs are failing, issue opened upstream: https://github.com/dependabot/dependabot-core/issues/8593

caleblloyd:
It's working again now!

PR #852, opened 2024-01-15
Bump NATS server to 2.10.9-alpine in /helm/charts/nats

svenfoo:
Update to the latest NATS server release.


svenfoo:
Please consider to also merge https://github.com/nats-io/k8s/pull/849

caleblloyd:
Thanks.  Dependabot normally does this for us but it is broken right now #840 :frowning:

PR #853, opened 2024-01-16
Nats-Kafka Service Account Support

ChrisDizenzo1047:
# Description

The Nats-Kafka `_helpers.tpl` already had the service account setup, but it seemed incomplete:

```
{{/*
Create the name of the service account to use
*/}}
{{- define "nats-kafka.serviceAccountName" -}}
{{- if .Values.serviceAccount.create }}
{{- default (include "nats-kafka.fullname" .) .Values.serviceAccount.name }}
{{- else }}
{{- default "default" .Values.serviceAccount.name }}
{{- end }}
{{- end }}
```

There was no usage of `nats-kafka.serviceAccountName`, so this PR just adds the usage of that. If create is true it creates a service account, if not it uses the value from serviceAccount.name.

# Testing

Tested the helm chart from our internal branch and worked as intended.

### DIFF
--- a/helm/charts/nats-kafka/Chart.yaml
+++ b/helm/charts/nats-kafka/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 0.15.3
+version: 0.15.4
 appVersion: 1.4.2
 type: application
 name: nats-kafka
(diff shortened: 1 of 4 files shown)


wallyqs (approved):
LGTM

PR #857, opened 2024-01-18
revert erroneuos dependabot bump of nats-server-config-reloader

### LINKED ISSUE #856: server-config-reloader has a non existing tag in the values.yaml of the helm chart
### What version were you using?

2.10.9

### What environment was the server running in?

kubernetes helm chart

### Is this defect reproducible?

yes it is reproducible, install using helm on Kubernetes and this should prevent the pods from starting up.

### Given the capability you are leveraging, describe your expectation?

I expect the values.yaml to contain a tag for the server-config-reloader that is actually available

### Given the expectation, what is the defect you are observing?

The helm chart is currently unable to run using the default values.yaml.

caleblloyd:
Resolves #856

This change was un-released so we should not need to bump the chart version

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -353,7 +353,7 @@ reloader:
   enabled: true
   image:
     repository: natsio/nats-server-config-reloader
-    tag: 0.14.2
+    tag: 0.14.1
     pullPolicy:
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

PR #862, opened 2024-02-05
Bump NATS server to 2.10.10-alpine in /helm/charts/nats

svenfoo:
Might make sense to try to restore the `dependabot` setup that used to do this until it stopped working at some point end of last year. For now, here's another PR to do the update manually.


svenfoo:
Please also merge https://github.com/nats-io/k8s/pull/855

caleblloyd:
Dependabot works.  It runs `daily` which means:

> Runs on every weekday, Monday to Friday.

I manually triggered it (would have run automatically later today) and it opened #863

Issue #772, opened 2023-08-02
[nats helm] Support config for `gateway.gateways` and `leafnodes.remotes`


caleblloyd:
`gateway.gateways` and `leafnodes.remotes` are both lists that support a `tls` block

Also, a server might not need to be listening on the `leafnodes` port in order to have a `leafnodes.remote` block.  The same may be true of a gateway.  If this is the case, these should probably be keys under `config` such as `config.leafnodeRemotes` and `config.gatewayRemotes` so that the Leafnodes/Gateways Servers can be managed separately from Clients

SqiSch:
Try with:
`  leafnodes:
    enabled: true
    merge:
      remotes:
        - url: nats://nats.infra:7422`

J11522:
Any updates on this?
We are currently running into trouble configuring the remotes for the leafnodes when it comes to the cluster credentials.
Having dedicated keys would greatly help, e.g. mounting the credentials file from a secret is (AFAIK) currently not possible.

Alternatively this could be solved using generic extraVolume/extraVolumeMounts pattern.
If you feel this is something that could be done, I can create a dedicated ticket and potentially provide input

mxchinegod:
This is a drastically needed feature. If it cannot be added by tomorrow I will do it.

mxchinegod:
This is a much needed feature

mccullough-ea:
I had to hard code the ca_file path for our leafnodes for each of the remotes that offer certs signed by our private CA. 
`tls:
   ca_file: /etc/nats-ca-cert/ca.crt`

I would love the remotes to be able to inherit the settings from tlsCA like all the other sections.

tommyjcarpenter:
Adding my support for this. I raised an issue that was seemingly a duplicate before finding this.

PR #870, opened 2024-02-21
🔮 feat(wip): extraVolume & mounts

mxchinegod:
https://github.com/nats-io/k8s/issues/772


caleblloyd:
`extraVolumes` and `extraVolumeMounts` are not necessary.  They can already be achieved with:


```yaml
podTemplate:
  patch:
  - op: add
    path: /spec/volumes/-
    value:
      name: my-secret
      secret:
        secretName: my-secret

container:
  patch:
  - op: add
    path: /volumeMounts/-
    value:
      name: my-secret
      mountPath: /etc/my-secret
```

PR #874, opened 2024-03-07
Add parameter for configuring the Jetstream domain

protochron:
Add the ability to set a Jetstream domain on the NATS cluster.


caleblloyd:
If the value only effects 1 place in the config, as this one does, we prefer to use the `merge` key under the respective config section.  That way we don't have to add to `values.yaml` for every possible config option, and any valid NATS Config is already supported!  Example:

```
config:
  jetstream:
    merge:
      domain: your-domain
```

Full docs with all possible sections that can be merged: https://artifacthub.io/packages/helm/nats/nats

PR #879, opened 2024-03-20
[nats helm] add chart-specific value support for imagePullSecrets

walsm232:
This commit takes care of allowing a user to explicitly set imagePullSecrets at the NATS and NATS Box level instead of needing to do it via a global value.

Supporting this like it is currently via only a **global** value could be problematic for a couple of reasons when this NATS chart is a dependency of another chart:
- when you define **imagePullSecrets** at the global level, it affects all dependencies using that Helm value. This may not be desirable if you have multiple dependencies, each needing different image pull secrets.
- users might need different image pull secrets for different dependencies or even within different components of the same dependency. By only having a global override, this fine-grained control is limited.
- managing **imagePullSecrets** for different dependencies solely through a global Helm value could lead to complexity and potential confusion.

### DIFF
--- a/helm/charts/nats/files/nats-box/deployment/pod-template.yaml
+++ b/helm/charts/nats/files/nats-box/deployment/pod-template.yaml
@@ -10,9 +10,10 @@ spec:
   # service discovery uses DNS; don't need service env vars
   enableServiceLinks: false
   
-  {{- with .Values.global.image.pullSecretNames }}
+  {{- $imagePullSecrets := default .Values.global.image.pullSecretNames .Values.natsBox.container.image.pullSecretNames }}
+  {{- if $imagePullSecrets }}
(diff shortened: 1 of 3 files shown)


caleblloyd:
First off, thank you for contributing.  I see your point, but we try to add as little additional schema as possible to this chart. 

hen a property only affects a single resource, we prefer to merge/patch it into place using the same schema as it's k8s resource.  This should be achievable today using:

```
# merge a pull secret for nats pods, this will override the global setting
podTemplate:
  merge:
    spec:
      imagePullSecrets:
      - name: regcred

# merge a pull secret for natsBox pods, this will override the global setting
natsBox:
  podTemplate:
    merge:
      spec:
        imagePullSecrets:
        - name: regcred
```

PR #882, opened 2024-03-26
fix nats-box deployment so nats user can be used

ddanat-smm:
When installing nats-box with the helm chart the current implementation saves the current directory as the "work_dir". In the Dockerfile that is "/root". Even though the user "nats" with uid 1000 is created in the Dockerfile, if the current deployment (which just does a sleep infinity) starts as uid 1000, work_dir gets saved as "/root" and when the cd $work_dir comes, the container crashes. This fixes that issue.

Sorry if this doesn't quite adhere to your PR guidelines - this is just a quick and dirty commit

### DIFF
--- a/helm/charts/nats/files/nats-box/deployment/container.yaml
+++ b/helm/charts/nats/files/nats-box/deployment/container.yaml
@@ -10,7 +10,6 @@ command:
 - sh
 - -ec
 - |
-  work_dir="$(pwd)"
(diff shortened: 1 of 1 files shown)


caleblloyd:
I think a better fix may be changing the exec to:

```
exec /entrypoint.sh "$@"
```

Because the [nats-box entrypoint already handles this logic](https://github.com/nats-io/nats-box/blob/main/entrypoint.sh)

That will also not break anyone who is using `workingDir`

PR #883, opened 2024-04-01
[nats helm] call natsbox entrypoint

caleblloyd:
Same as #882 except calls natsbox entrypoint, which figures out if `$HOME` should be changed into

### DIFF
--- a/helm/charts/nats/files/nats-box/deployment/container.yaml
+++ b/helm/charts/nats/files/nats-box/deployment/container.yaml
@@ -22,7 +22,8 @@ command:
   fi
   {{- end }}
   cd "$work_dir"
-  exec /entrypoint.sh "$0" "$@"
+  exec /entrypoint.sh "$@"
+- --
 args:
 - sh
 - -ec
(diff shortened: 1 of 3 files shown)


helm/charts/nats/files/nats-box/deployment/container.yaml
```diff
@@ -22,8 +22,10 @@ command:
   fi
   {{- end }}
   cd "$work_dir"
-  exec sh -ec "$0"
+  exec /entrypoint.sh "$0" "$@"
```
caleblloyd, line 25: @samuelattwood is there an equivalent single command for `"$0" "$@"` that you know of?

samuelattwood: You should be able to use `"${@:0}"`

caleblloyd: ugh I think that may be `bash` only and not `sh`

caleblloyd: Could pass a dummy arg for $0, like `--`, and skip it?

samuelattwood (approved):
LGTM

PR #901, opened 2024-05-23
Update README.md

runeb-wombo:
Erroneous example config for jetStream. 

memoryStore mentioned twice in sample config. Second mention should be for fileStore.


caleblloyd (approved):
Thanks! LGTM

PR #912, opened 2024-06-17
Don't enable WebSocket compression by default

neilalexander:
Currently there are some problems where the flate writer might take up more allocations than expected when there are lots of WebSocket clients.

Signed-off-by: Neil Twigg <neil@nats.io>

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
 - nats
 - messaging
 - cncf
-version: 1.1.12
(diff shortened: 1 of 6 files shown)


caleblloyd:
My original thought was that enabling compression by default was good for most common use cases, since Cloud Provider bandwidth is so expensive.

But if the desire is to optimize for high performance scenarios on fast network connections, I digress.

The tests that assert compression is on in the chart’s `test` sub directory need to be changed from true to false

caleblloyd (approved):

PR #971, opened 2025-02-26
Add support for specifying image digest and full image name in NATS

jereml99:
This PR adds two new Helm chart values: digest and fullname. The image resolution now follows this priority: tag < digest < fullname, ensuring immutable image references for improved security and compliance. 

issue: #970

PR #972, opened 2025-02-27
Bump NATS chart version to 1.3.0

jereml99:
Bump NATS chart version to 1.3.0 after following changes:

https://github.com/nats-io/k8s/pull/971


svenfoo:
As far as I can see this should have been merged before making the `nats-1.2.11` release, because #971 introduced new Helm values.

PR #973, opened 2025-02-27
add extraVolumes and extraVolumeMounts for surveyor

spencerharmon:
Our use case is to mount ca certs into the surveyor container, but this will enable arbitrary volume mounts which might be useful to someone. I could also write something more opinionated for the CA case if that's preferred. Is this something yall would support?


wallyqs (approved):
LGTM!

PR #979, opened 2025-03-12
[NACK] Add custom labels to ClusterRole or Role

### LINKED ISSUE #978: Add Labels to RBAC Jetstream Controller Role
### What motivated this proposal?

We use a CICD user in our kubernetes cluster to automate deployments.
This CICD user has a namespaced `admin` role attached to it.

The `jetstream-controller-role` role doesn't have the `rbac.authorization.k8s.io/aggregate-to-admin: "true"` label on it, which means we would have to either manually modify the role with the label or attach the role to our CICD user.

### What is the proposed change?

Add the ability to add custom labels to the `jetstream-controller-role` role to be able to set `rbac.authorization.k8s.io/aggregate-to-admin: "true"`

### Who benefits from this change?

_No response_

### What alternatives have you evaluated?

* Manually modifying the `jetstream-controller-role` role
* Manually attaching the role to users

rblaine95:
Closes #978 

Add ability to add custom labels (e.g: `rbac.authorization.k8s.io/aggregate-to-admin="true"`) to NACK ClusterRole or Role


helm/charts/nack/templates/rbac-jetstream-controller.yml
```diff
@@ -13,6 +13,8 @@ kind: {{ $kind }}
 metadata:
   name: {{ template "jsc.serviceAccountName" . }}-{{ $kindSuffix }}
   namespace: {{ include "jsc.namespace" . }}
+  labels:
+    {{- tpl (toYaml .Values.rbacLabels) . | nindent 4 }}
```
kejne, line 17: This will cause a templating error if no label is passed in. Use an if statement in combination to make sure that it works when not passing any value.
Similar as for `podAnnotations`.

kejne (commented):
(I am not a maintainer, just contributing with some nearby stuff)

PR #987, opened 2025-04-08
fix(charts/stan): remove -channelz, -serverz from exporter.args default

hatant39:
Both `GetStreamingChannelz` and `GetStreamingServerz` were removed from `NATSExporterOptions` in https://github.com/nats-io/prometheus-nats-exporter/pull/329 due to the `nats-streaming-server` deprecation.   Since `.Values.exporter.image` defaults to `latest`, new pods would enter a `CrashLoopBackOff` with the metrics container failing to start due to the undefined args.  Example errors below:

```
Defaulted container "metrics" out of: metrics, stan
flag provided but not defined: -channelz
Usage of /prometheus-nats-exporter:
…
```

```
Defaulted container "metrics" out of: metrics, stan
flag provided but not defined: -serverz
Usage of /prometheus-nats-exporter:
…
```
This PR only removes the `-channelz` and `-serverz` streaming args from the default case when the exporter is enabled and `.Values.exporter.args[]` is empty.

### DIFF
--- a/helm/charts/stan/Chart.yaml
+++ b/helm/charts/stan/Chart.yaml
@@ -11,7 +11,7 @@ keywords:
 - replay
 - statefulset
 - cncf
-version: 0.13.0
+version: 0.13.1
 maintainers:
 - email: info@nats.io
   name: The NATS Authors
--- a/helm/charts/stan/templates/statefulset.yaml
+++ b/helm/charts/stan/templates/statefulset.yaml
@@ -314,8 +314,6 @@ spec:
           - -routez
           - -subz
           - -varz
-          - -channelz
-          - -serverz
           - http://localhost:8222/
           {{- end }}
           ports:


wallyqs (approved):
LGTM

ripienaar:
I think we need to remove the `stan` chart instead.   The service was deprecated almost 2 years ago, has known security issues and we really need to signal in the strongest possible manner that continued use is not ok.

hatant39:
> I think we need to remove the `stan` chart instead. The service was deprecated almost 2 years ago, has known security issues and we really need to signal in the strongest possible manner that continued use is not ok.

@ripienaar I tend to agree.  Just wanted sure about the best path forward.

wallyqs:
removing sounds good to me too

ripienaar:
What's the best way to go about that @wallyqs and maybe @samuelattwood ?

samuelattwood:
Yes, I was planning on removing this chart.

hatant39:
> @hatant39 this was fixed in #988 Let us know if it still broken for you

Thanks for the update @alexbozhenko!  We're good.  Glad to see the `stan` chart removed.

PR #988, opened 2025-04-09
Remove stan

alexbozhenko:
https://github.com/nats-io/prometheus-nats-exporter/pull/329 removed flags that were used for stan.
It causes problems:
https://github.com/nats-io/k8s/pull/987

Bump prometheus-nats-exporter version and remove stan chart.


helm/charts/nack/values.yaml
```diff
@@ -151,7 +151,7 @@ tolerations: []
 #    value: "value"
 #    effect: "NoSchedule|PreferNoSchedule|NoExecute(1.6 only)"
 
-# Resource requests and limits for primary stan container
+# Resource requests and limits for primary nats container
```
samuelattwood, line 154: `stan` was wrong too, but it's not really a `nats` container. Should probably just say `nack` or `jetstream-controller`

samuelattwood (approved):

Issue #785, opened 2023-08-20
Prom-Exporter container failing : [ERR] Could not find server_id: invalid character 'C' looking for beginning of value


ksingh-scogo:
After deploying nats cluster using helm the `prom-exporter` container is failing  inside the jestream pod, with the error  `[ERR] Could not find server_id: invalid character 'C' looking for beginning of value` 

This is causing the health check to fail with error `  Warning  Unhealthy               75s (x16 over 3m45s)  kubelet                  Startup probe failed: HTTP probe failed with statuscode: 400`

- Versions
```
- Kubernetes Version: v1.26.6
- NATS Helm Chart Version: 1.0.2
- NATS App Version: v2.9.21
```
- `kubectl get all`
```
NAME                                      READY   STATUS    RESTARTS   AGE
pod/nats-jetstream-0                      2/3     Running   0          13m
pod/nats-jetstream-1                      2/3     Running   0          13m
pod/nats-jetstream-2                      2/3     Running   0          13m
pod/nats-jetstream-box-56586bdf9f-phtzs   1/1     Running   0          13m

NAME                              TYPE           CLUSTER-IP     EXTERNAL-IP     PORT(S)                               AGE
service/nats-jetstream            LoadBalancer   10.0.111.248   20.219.34.176   4222:31806/TCP,8080:30823/TCP         13m
service/nats-jetstream-headless   ClusterIP      None           <none>          4222/TCP,8080/TCP,6222/TCP,8222/TCP   13m

NAME                                 READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/nats-jetstream-box   1/1     1            1           13m

NAME                                            DESIRED   CURRENT   READY   AGE
replicaset.apps/nats-jetstream-box-56586bdf9f   1         1         1       13m

NAME                              READY   AGE
statefulset.apps/nats-jetstream   0/3     13m
```

- `values.yaml`
```
################################################################################
# Global options
################################################################################
global:
  image:
    # global image pull policy to use for all container images in the chart
    # can be overridden by individual image pullPolicy
    pullPolicy: IfNotPresent
    # global registry to use for all container images in the chart
    # can be overridden by individual image registry
    registry:

  # global labels will be applied to all resources deployed by the chart
  labels:
    app.kubernetes.io/name: nats
    app.kubernetes.io/version: 2.9.21
    app.kubernetes.io/managed-by: Helm

################################################################################
# Common options
################################################################################
# override name of the chart
nameOverride:
[...]

PR #991, opened 2025-04-10
Fix PromExporter with TLS enabled Monitor

swwolf:
This PR fixes an issue that occurs when TLS is enabled for the Monitor and the Prometheus Exporter is also enabled. A hardcoded `"http"` value prevents the `promExporter` container from starting up healthily.

To resolve this, the protocol used to query the Monitor (either `http` or `https`) is now selected automatically based on the TLS setting. Additionally, the domain that Prometheus Exporter queries is now configurable.

Note that the domain cannot be set to `localhost` if the TLS certificate used by the Monitor does not include `localhost` as a Subject Alternative Name (SAN). In such cases, `monitorDomain` must match a name present in the certificate.

See https://github.com/nats-io/k8s/issues/785


wallyqs (approved):
LGTM

PR #1016, opened 2025-06-10
refactor(nack): Use AppVersion to update to NACK image 0.18.2

kejne:
- Updates NACK chart to NACK controller 0.18.2
- Updates CRD changes introduced in NACK 0.18.0
- Helm chart refactored to use built in `appVersion` for it's intended use
  - Still possible to override the image tag as previously


kejne:
@wallyqs could we get this merged soon?
Would be great to get the new release out as well as well needed fixes. :pray:

kejne:
@Jarema possible to also merge and release?

Issue #905, opened 2024-05-31
Modify the surveyor chart so that it can connect using the parameters --user --password


MilkyWay-core:
### What motivated this proposal?

We using helm secrets and argoCD vault plugin. We need the ability to encrypt the password in the parameters separately from the entire string
nats://login:pass@server:port doesn't work for us

### What is the proposed change?

Change the chart

deployment.yaml

`          args:
           {{- with .Values.config.basicCredentials }}
            - --user={{ .user }}
           {{- if .password }}
            - --password={{ .password }}
           {{- end }}
           {{- end }}
`

values.yaml

`
config:
  # Required if NATS basic auth is enabled
  # basicCredentials:
  #   user: admin
  #   password: password
`

### Who benefits from this change?

The correct way to transfer credentials will avoid storing the password in clear text in the values.yaml

### What alternatives have you evaluated?

-

francescov1:
@MilkyWay-core made a PR for this https://github.com/nats-io/k8s/pull/1019

PR #1019, opened 2025-06-24
francescov1/add user password auth to surveyor

francescov1:
This PR adds support for `--user` and `--password` to the Surveyor Helm chart. It uses a secret for the password.

Addresses #905

### DIFF
--- a/helm/charts/surveyor/templates/deployment.yaml
+++ b/helm/charts/surveyor/templates/deployment.yaml
@@ -48,14 +48,6 @@ spec:
             - --nkey=/nkey/{{ .secret.key }}
            {{- end }}
 
-           {{- with .user }}
-            - --user={{ . }}
-           {{- end }}
-
-           {{- with .password }}
-            - --password=$(NATS_ADMIN_PASSWORD)
-           {{- end }}
-
            {{- with .servers }}
             - -s={{ . }}
            {{- end }}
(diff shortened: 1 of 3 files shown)


helm/charts/surveyor/templates/deployment.yaml
```diff
@@ -48,6 +48,14 @@ spec:
             - --nkey=/nkey/{{ .secret.key }}
            {{- end }}
 
+           {{- with .user }}
+            - --user={{ . }}
+           {{- end }}
+
+           {{- with .password }}
+            - --password=$(NATS_ADMIN_PASSWORD)
```
alexbozhenko, line 56: I am not sure if this a real concern, but this password will show up in the process list.
Maybe better to pass `NATS_SURVEYOR_PASSWORD` env var to the process?

wallyqs: yes this is better handled as an env var, though also you do not need these since you can set the string from nats as `nats://user:pass@endpoint:4222` and that would use those creds

wallyqs: maybe env var based concatenation could be an alternative?
```
env: 
        - name: NATS_URL
          value: $(NATS_USER):$(NATS_PASSWORD)@$(NATS_ENDPOINT):4222
```

wallyqs (approved):
LGTM

PR #1032, opened 2025-08-08
Feature/add crd nats 0.19.0

ksemele:
We have a strange situation here.
An updated CRD for new NACK release need to copy to k8s repo...
I think we can do this better in the future, but for now I just copied new CRD from here: https://github.com/nats-io/nack/blob/main/deploy/crds.yml

### DIFF
--- a/helm/charts/nack/Chart.yaml
+++ b/helm/charts/nack/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nack
 version: 0.29.0
-appVersion: "0.18.2"
+appVersion: "0.19.0"
 description: A Helm chart for NACK - NAts Controller for Kubernetes
 keywords:
 - nats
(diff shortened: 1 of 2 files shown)


helm/charts/nack/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nack
-version: 0.29.0
-appVersion: "0.18.2"
+version: 0.29.1
+appVersion: "0.19.0"
```
svenfoo, line 4: You may want to bump to "0.19.1" now.

helm/charts/nack/crds/crds.yml
```diff
@@ -232,6 +232,14 @@ spec:
...
                 type: boolean
                 default: false
+              allowMsgTtl:
+                description: When true, allows header initiated per-message TTLs. If disabled, then the `NATS-TTL` header will be ignored.
+                type: boolean
+                default: false
+              subjectDeleteMarkerTtl:
+                description: Enables and sets a duration for adding server markers for delete, purge and max age limits.
+                type: string
+                default: ''
```
svenfoo, line 242: Nice! This should fix the warnings that we are seeing lately.

svenfoo (approved):

PR #1034, opened 2025-08-13
make the nats format config helper for environment variable match as …

omer9564:
…few times as possible to allow having two env vars in same row

When using [variables](https://github.com/nats-io/k8s/pull/1034) the chart currently requires a << $name >> format as mentioned [here](https://github.com/nats-io/k8s/tree/main/helm/charts/nats#nats-config-units-and-variables).
The problem is that if you want to create a values like
```
<< $SYS_ACCOUNT_ID >>: << $SYS_ACCOUNT_JWT >>
```
the template generates
```
$SYS_ACCOUNT_ID >>": "<< $SYS_ACCOUNT_JWT
```

expected to get:
```
$SYS_ACCOUNT_ID: $SYS_ACCOUNT_JWT
```

### DIFF
--- a/helm/charts/nats/templates/_helpers.tpl
+++ b/helm/charts/nats/templates/_helpers.tpl
@@ -282,7 +282,7 @@ output: string with following format rules
 */}}
 {{- define "nats.formatConfig" -}}
   {{-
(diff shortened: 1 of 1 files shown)


johnweldon (approved):
LGTM

PR #1041, opened 2025-08-22
fix: add missing fields

carlosjgp:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#statefulsetspec-v1-apps 
volumeClaimTemplates ==>
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#persistentvolumeclaim-v1-core

### DIFF
--- a/helm/charts/nats/files/stateful-set/resolver-pvc.yaml
+++ b/helm/charts/nats/files/stateful-set/resolver-pvc.yaml
@@ -1,4 +1,6 @@
 {{- with .Values.config.resolver.pvc }}
+apiVersion: v1
+kind: PersistentVolumeClaim
 metadata:
   name: {{ .name }}
 spec:


carlosjgp:
ksnip_20250903-140859

carlosjgp:
> Hi @carlosjgp. Would you mind explaining the motivation for this change? Defining `apiVersion` and `kind` is not required in the context of a statefulset `volumeClaimTemplates` array. The [Kubernetes docs](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components) follow the same pattern as the NATS chart.

While these fields are optional, after creating the object the K8s/k3s API returns these optionial fields as part of the spec

This causes constant differences in GitOps tools, in my case ArgoCD.

if we don't merge this PR then every user of this chart would have to add an exclusion to to their gitops tool to explicitly ignore this difference... this is normally not recommended since other expected changes could be ignore too

I hope that makes sense

Also that documentation is 9y old
ksnip_20250904-144638

PR #1067, opened 2025-10-09
feat: add missing resources declaration for containers

alxyok:
Adds a resources declaration block for setting requests and limits for containers in:
* `nats.nats-box`
* `nats.stateful-set`
* `nats-account-server`


wallyqs (approved):
LGTM

PR #1078, opened 2025-10-24
Bump NATS Server to 2.12.1

samuelattwood:
First release in the NATS Server `2.12` line. `2.11` releases will remain on the `1.3` chart versions for consistency, but going forward we will align `major.minor` with the app version to allow better clarity with multiple supported minor versions.

For example:
`helm upgrade -i nats nats/nats --version "~2.12"`

### DIFF
--- a/helm/charts/nats/Chart.yaml
+++ b/helm/charts/nats/Chart.yaml
@@ -1,12 +1,12 @@
 apiVersion: v2
-appVersion: 2.11.10
+appVersion: 2.12.1
 description: A Helm chart for the NATS.io High Speed Cloud Native Distributed Communications Technology.
 name: nats
 keywords:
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

samuelattwood:
@alexbozhenko We have been unable to replicate any metaleader issues like mentioned in #1061. We held off on updating the chart until `2.12.1` and can investigate further if we get more details from the user.

wrenix:
What is the breaking change of this commit? There is no changelog why this helmchart was increase by a major Version ... Helm Charts should be in semantic versioning.

kejne:
Thanks for the clarification @samuelattwood . That makes sense! 👌

PR #1104, opened 2025-12-22
fix: add api version kind to statefulset pvc

### LINKED ISSUE #1103: volumeClaimTemplates triggers persistent Out-of-Sync status in GitOps (ArgoCD)
### What version were you using?

NATS Helm Chart 2.12.3

### What environment was the server running in?

kind via ArgoCD

### Is this defect reproducible?

```
helm template ./helm/charts/nats --show-only templates/stateful-set.yaml --set config.jetstream.enabled=true
```

### Given the capability you are leveraging, describe your expectation?

```
  volumeClaimTemplates:
  - apiVersion: v1
    kind: PersistentVolumeClaim
    metadata:
      name: release-name-nats-js
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 10Gi
```

### Given the expectation, what is the defect you are observing?

```
  volumeClaimTemplates:
  - metadata:
      name: release-name-nats-js
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 10Gi
```

k8s will helpfully add the missing `apiVersion` & `kind` upon admission, but causes a drift between what is applied and what is deployed

arcanez:
this fixes #1103 

*Problem*: When using ArgoCD, the NATS StatefulSet always shows a "diff" for `volumeClaimTemplates`. This is because the chart omits `apiVersion` and `kind`, which the Kubernetes API server then automatically adds to the live resource. ArgoCD interprets this as a drift from the desired state.

*Solution*: Explicitly add `apiVersion: v1` and `kind: PersistentVolumeClaim` to the `jetstream-pvc.yaml` and `resolver-pvc.yaml` files.

*Impact*: This is a non-breaking change that ensures the rendered Helm output matches the live cluster state, resulting in a "Healthy/Synced" status in GitOps controllers. 
*CAVEAT*: This may result in a sync error for existing deployments since `volumeClaimTemplates` is immutable, but a delete & recreate of the `StatefulSet` can remedy this.

### DIFF
--- a/helm/charts/nats/files/stateful-set/jetstream-pvc.yaml
+++ b/helm/charts/nats/files/stateful-set/jetstream-pvc.yaml
@@ -1,4 +1,6 @@
 {{- with .Values.config.jetstream.fileStore.pvc }}
+apiVersion: v1
+kind: PersistentVolumeClaim
 metadata:
   name: {{ .name }}
 spec:
--- a/helm/charts/nats/files/stateful-set/resolver-pvc.yaml
+++ b/helm/charts/nats/files/stateful-set/resolver-pvc.yaml
@@ -1,4 +1,6 @@
 {{- with .Values.config.resolver.pvc }}
+apiVersion: v1
+kind: PersistentVolumeClaim
 metadata:
   name: {{ .name }}
 spec:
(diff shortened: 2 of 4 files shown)


svenfoo (approved):
lgtm

PR #1108, opened 2026-01-09
Add option to re-enable nats advertisements

wallyqs:


helm/charts/nats/values.yaml
```diff
@@ -53,6 +53,9 @@ config:
     port: 6222
     # must be 2 or higher when jetstream is enabled
     replicas: 3
+    # set to false to allow cluster nodes to advertise their addresses
+    # so that clients can reconnect without extra DNS lookups.
```
wallyqs:
```suggestion
    # so that clients can reconnect without extra DNS lookups.
    # Note: in case clients have external connectivity make sure to define the `advertise` section as well.
    # If clients are behind a load balancer it is best to leave this as is.
```

PR #1116, opened 2026-01-28
Add Release Branch Trigger for 2.11 Updates

samuelattwood:
- Also updated to fix Zizmor-identified concerns

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,34 +5,34 @@ updates:
   schedule:
     interval: "weekly"
   cooldown:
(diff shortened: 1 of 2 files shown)


wallyqs:
what is 2.11, the natscli?

samuelattwood:
@wallyqs This is to track updates for `nats-server` `2.11.X` so main can stay on the latest `2.12` but we can simultaneously update the older version of the chart when we have `2.11.` patch releases.

wallyqs (approved):

PR #1148, opened 2026-03-27
Clarify configChecksumAnnotation behavior for non-reloadable params

jtnats:
## Purpose
Clarify configChecksumAnnotation behavior for non-reloadable params

## Summary
- Expand the `configChecksumAnnotation` comment in values.yaml to warn that non-reloadable parameters (e.g. `max_mem`, `max_file`) are silently ignored when the reloader sends SIGHUP
- Recommend setting `configChecksumAnnotation: true` when using non-reloadable parameters so that a rolling restart applies them

## Motivation
With the default `configChecksumAnnotation: false`, the reloader signals `SIGHUP` on config changes. For non-reloadable parameters, the server rejects the reload and the new values never take effect but there is also no visible error to the user. This has been a source of confusion.

### DIFF
--- a/helm/charts/nats/values.yaml
+++ b/helm/charts/nats/values.yaml
@@ -482,7 +482,18 @@ statefulSet:
 podTemplate:
   # adds a hash of the ConfigMap as a pod annotation
   # this will cause the StatefulSet to roll when the ConfigMap is updated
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
lgtm, maybe file a separate server issue that those fields are not reloadable and put
in the comment

PR #1183, opened 2026-08-05
Address review feedback on leafnode remotesHelm chart values

jtnats:
## Overview

This PR addresses all findings from the review of the leafnode remotes feature: both blocking issues (the no-op credentials reloader watch and the global `tlsCA` never reaching remotes), both follow-ups (the list-recursion side effect and the silent credentials misconfiguration), and the three nits. One reviewer suggestion is implemented differently than proposed, with rationale below.

## Changes

The per-remote `tls` block now requires `enabled: true`, matching every other tls block in the chart, and a bare `tls: {enabled: true}` applies the global `tlsCA` as `ca_file` — the private-CA hub case that previously rendered nothing. The review suggested rendering the block automatically whenever `tlsCA` is enabled, but that is not safe: `server/leafnode.go:3508` sets `tlsRequired := remote.TLS || remote.TLSConfig != nil`, so any rendered `tls` block, even one carrying only `ca_file`, forces a TLS handshake and would break plain `nats://` remotes; and since `ca_file` replaces the system root pool, injecting the private CA would also break remotes that rely on public CAs, such as NGS. The two cases cannot be distinguished automatically, so opting in per remote is the only correct general design, and it resolves the tls-block asymmetry nit as a side effect.

The `credentials` case is removed from `nats.reloaderConfig`. As the review demonstrated live and the source confirms, the server reads the creds file from disk on every connection attempt and a reload signal neither re-reads credentials nor affects established leafnode connections, so the watch could only ever send a signal the server ignores while logging as if rotation had been applied. Rotation semantics are now documented as next-reconnect in the README and values.yaml. The certificate watches remain, because remote `TLSConfig` is only rebuilt on reload, so those watches do real work for future reconnect attempts.

A `credentials` block without `secretName` now fails the render with `config.leafnodes.remotes credentials must set secretName`, matching the existing url guard, instead of silently emitting a remote that connects anonymously. Hand-mounted creds paths remain reachable through `config.leafnodes.patch`. The url-plus-urls combination is now documented in values.yaml and the README, and the stray double blank line in the README is fixed.

## Behavior

Default rendered output remains byte-identical to main, with one intentional and now-documented exception: configs that already reference certificate files inside config lists, for example `gateway.merge.gateways[].tls`, gain reloader watch args from the list recursion, which mutates the pod spec and causes one rolling restart on upgrade. Within this feature, a per-remote `tls` block that previously set only `secretName` must now also set `enabled: true` to render.

## Testing

Added `TestConfigLeafnodeRemoteTlsCA` rendering the exact missing-`ca_file` scenario from the review, `TestConfigLeafnodeRemotesRenderFail` covering both render guards through a new `HelmRenderError` harness helper, and `TestReloaderTlsInConfigList` pinning the list-recursion behavior with a gateway merge fixture. `TestConfigLeafnodeRemotes` is updated for the `enabled` flag and the removed creds watch. The full suite passes, the nindent check and `helm lint` are clean, and the default-render diff against main was verified empty both with leafnodes disabled and enabled with no remotes.

### DIFF
--- a/helm/charts/nats/README.md
+++ b/helm/charts/nats/README.md
@@ -215,7 +215,7 @@ config:
 ### Leafnode Remotes and Synadia Cloud (NGS)
 
 `config.leafnodes.remotes` solicits [leafnode](https://docs.nats.io/running-a-nats-service/configuration/leafnodes) connections to remote NATS servers, with credentials mounted from an existing Kubernetes Secret.
-The config reloader watches the mounted credentials and certificate files, so rotating the Secret hot-reloads the server.
+Rotating a mounted credentials or certificate Secret takes effect on the next connection attempt to the remote, not on established connections: the server reads the credentials file fresh on every attempt, and the config reloader re-applies changed certificate files for future attempts.
 
 Example - connect to [Synadia Cloud](https://docs.synadia.com/cloud) (NGS):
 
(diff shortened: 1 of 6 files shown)


jtnats:
Ignore this PR, it's not ready.

PR #1194, opened 2026-09-17
Bump nats to v2.15.0

1995parham:
Bumps the `nats` chart to the [v2.15.0](https://github.com/nats-io/nats-server/releases/tag/v2.15.0) server release: chart `version` and `appVersion` kept in sync, and the default `container.image.tag` moved to `2.15.0-alpine`.

`go test` in `helm/charts/nats/test` passes locally.

One caveat: `nats:2.15.0-alpine` is not on Docker Hub yet — v2.15.0 was tagged only a few hours ago, and 2.14.7 took around 14 hours to show up as an image — so `ct install` will fail until the official image is published. Feel free to hold this until then; I can rebase whenever it suits you.

Unrelated question while I am here: there is no released chart for 2.14.7, and the [2.15 upgrade guide](https://docs.nats.io/release-notes/upgrade-to-2.15) recommends upgrading from at least 2.14.7 for the desired-state metalayer changes. That leaves chart users on 2.14.6 with no released intermediate step. Would you like a 2.14.7 chart released as well? I am happy to open that PR too.


tcarpenter-eqix:
Can we please get this released, we need the Jetstream fixes included in this 🙏

Jarema (approved):
LGTM!

